PR #4, opened 2015-01-06
Start bar charts at 0 on the y-axis

tmm1:


tmm1:
This generally makes it easier to compare bars side-by-side, but given that the scale is logarithmic too it might not matter.

tgxworld:
@tmm1 Seems like `min` doesn't work well with `type: logarithmic`. I'm ok with removing the logarithmic scale since the only reason why I set it in initially was to amplify the differences in the charts.

tgxworld:
I've gone ahead and [remove logarithmic scale](https://github.com/ruby-bench/ruby-bench-web/commit/53299d180903c439788bd5a82e4420058a6f94a7). Thanks for the feedback @tmm1 Do let me know if there are any more things we need to change/do.

Issue #25, opened 2015-02-16
Improve to click to commit's GitHub URL feature.


tgxworld:
http://community.rubybench.org/t/click-to-commit-feature/39

PR #26, opened 2015-02-16
Display range of commits that are untested when clicking on a point.

tgxworld:
https://github.com/ruby-bench/ruby-bench-web/issues/25


tgxworld:
@SamSaffron I've thrown up an initial implementation. Is this what you were looking for? I'm using GitHub's API here so rate limit is going to be an issue and I've not thought of a solution for it yet.

# Normal Range
screenshot from 2015-02-16 23 35 57

## Edge Cases
screenshot from 2015-02-16 23 36 07

tgxworld:
bloggif_54e2115a318a1

tgxworld:
@SamSaffron Instead of pulling all the commit details from GitHub's API, I decided to just link to the repo's compare. I very much prefer this method as compared to the first. 
 
screenshot from 2015-02-17 10 40 47

SamSaffron:
lets play with both and see which one we like better

tgxworld:
Ok I'll push the second one up first to avoid having to deal with GitHub's rate limit.

PR #32, opened 2015-03-02
Invalidate benchmark results if benchmark changes

tgxworld:
TODO:

- [ ] BenchmarkType for Discourse uses the official Discourse URL repo. Need to update it otherwise our benchmarks will not be valid.

- [x] BenchmarkTypeDigestJob needs to handle invalidating the BenchmarkRuns too

- [ ] Don't display benchmark_runs that are not valid.

### DIFF
--- a/db/migrate/20150302075441_add_validity_to_benchmark_runs.rb
+++ b/db/migrate/20150302075441_add_validity_to_benchmark_runs.rb
@@ -1,0 +1,5 @@
+class AddValidityToBenchmarkRuns < ActiveRecord::Migration
+  def change
+    add_column :benchmark_runs, :validity, :boolean, default: true
+  end
+end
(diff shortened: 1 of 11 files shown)


db/migrate/20150302075441_add_validity_to_benchmark_runs.rb
```diff
@@ -0,0 +1,5 @@
+class AddValidityToBenchmarkRuns < ActiveRecord::Migration
+  def change
+    add_column :benchmark_runs, :validity, :boolean, default: true
```
tgxworld: Can't think of a better name. Wanted to use valid but it clashes with Rails

tgxworld:
Need to resolve checkbox 1 first. Otherwise, Discourse benchmarks will be invalidated.

db/migrate/20150302075441_add_validity_to_benchmark_runs.rb
tgxworld: TODO: change to still_valid

tgxworld:
Closing in preference of https://github.com/ruby-bench/ruby-bench-web/pull/111

PR #37, opened 2015-03-13
add source block to solve multiple primary source warning

farukaydin:
```bundle``` commands shows up a warning message like the following.
```
Warning: this Gemfile contains multiple primary sources. Using `source` more than once without
a block is a security risk, and may result in installing unexpected gems. To resolve this 
warning, use a block to indicate which gems should come from the secondary source. To upgrade
this warning to an error, run `bundle config disable_multisource true`.
```

PR solve this security fix, and disable warning message.


tgxworld:
:+1: Thank you1

Issue #42, opened 2015-03-24
Migrate to Sidekiq


tgxworld:

tgxworld:
Related: https://github.com/ruby-bench/ruby-bench-web/pull/43

tgxworld:
Close by https://github.com/ruby-bench/ruby-bench-web/pull/43.

PR #41, opened 2015-03-24
Add deployment templates.

tgxworld:

PR #43, opened 2015-03-26
Use Sidekiq for background processing.

tgxworld:
Only merge after https://github.com/ruby-bench/ruby-bench-web/pull/41.

Closes: https://github.com/ruby-bench/ruby-bench-web/issues/42


tgxworld:
Merging this in.

PR #50, opened 2015-04-13
Tweaks

kirs:
@tgxworld 

Note: `AS::Notifications` is generally used for another kind of tasks and using normal class instead of AS::Notifications makes it better testable here.


tgxworld:
@kirs Thanks! This is alot nicer :smile: Is this good for merge?

kirs:
I suppose so.

tgxworld:
Thanks @kirs!

PR #65, opened 2015-05-11
Update highcharts

yangshun:
- Use rubygems.org's highcharts gem instead of rails-assets.org's
- Update highcharts gem version to 4.1.5


yangshun:
Not required anymore.

PR #85, opened 2015-06-01
Split release bechmark graphs into two when two results exist

klgilbert:
To improve usability of the release graphs, especially for 'rails/rails/releases' and activerecord/postgres_scope, they need to be split up into separate graphs, but there isn't an easy way to do this from the way it works now.


tgxworld:
From the way it is designed right now, we can only split graphs by benchmark types. `app_uri` and `api_uri_memory`.

Results stored within each benchmark run should be displayed within the same graph. For instance, Discourse benchmarks have results for 25, 75, 90, 99 percentiles. Splitting them up into different graphs doesn't really make sense.

With all that said, I think we need that flexibility of being able to split the graphs up without having to create a new `BenchmarkType` for it. I'm not sure how to do it as of right now though...

cc/ @SamSaffron

klgilbert:
Yeah, you're right. It wouldn't make sense/wouldn't work for the Discourse benchmarks. This is probably something that needs to done with Javascript.

klgilbert:
It's a little hacky, but it does work. What do you think?

tgxworld:
I don't think this is the way we should go. Any new benchmarks that contains two columns will split. Could you explore a solution that involves changing the schema? I think we need BenchmarkType to have many forms of results.

klgilbert:
Ok

tgxworld:
Close by https://github.com/ruby-bench/ruby-bench-web/pull/108.

PR #86, opened 2015-06-02
Disable webkit appearance for select element

yangshun:
On Chrome, initially it was:
screen shot 2015-06-02 at 7 23 01 pm

I changed it to:
screen shot 2015-06-02 at 7 22 54 pm


tgxworld:
Hmm do you know if compass will help here? Otherwise, we have to handle all the different browser.

yangshun:
No la it's just webkit browsers, on FF it's fine (:

PR #89, opened 2015-06-12
Make footer sticky for screen sizes larger than 865px height

kennethlimjf:


tgxworld:
@kennethlimjf Thanks for contributing!! However, I'm not too sure what we are trying to fix here. Could you provide some screenshots of the before and after?

yangshun:
It's so that the footer will be at the bottom of the screen even though the site content isn't taller than the screen height. NUSMods does it too.

It's a good idea! But `flex` is not supported by IE9. @tgxworld do you want to support IE9? Haha.

Alternatively, there's a non-`flex` solution but involves hardcoding of the footer height.

yangshun:
I actually don't see a page that has so little content that the footer isn't at the bottom. Perhaps there isn't a need for a sticky footer.

app/assets/stylesheets/base.scss
```diff
@@ -1,3 +1,13 @@
 body {
   background-color: white;
 }
+
+.site {
```
yangshun: I feel that adding this styles directly to `body` is ok, no need for a `.site` class.

tgxworld:
> It's so that the footer will be at the bottom of the screen even though the site content isn't taller than the screen height. NUSMods does it too.
> 
> It's a good idea! But flex is not supported by IE9. @tgxworld do you want to support IE9? Haha.
> 
> Alternatively, there's a non-flex solution but involves hardcoding of the footer height.

screenshot from 2015-06-13 00 16 22

Ahh Ic. Lets move the web forward by dropping IE9 :trollface:

kennethlimjf:
Ah ya should have added a screenshot. I added the sticky footer since it appeared non-sticky on my 23" monitor. But for most laptop screens, it shouldn't be a problem.

But anyways, now it should look good on all screens with this PR :smile: (with the exception of IE9)

Updated the CSS not to use class.

tgxworld:
@yangshun I'll let you merge this if its ok :smile:

yangshun:
Just tried it. It works really well and elegant. Thanks!

tgxworld:
@kennethlimjf This is awesome thanks!

kennethlimjf:
Thanks :)

PR #91, opened 2015-06-17
Allow patterns to be passed to rails benchmarks

klgilbert:


tgxworld:
@klgilbert Tests are failing :smiley_cat:

app/jobs/remote_server_job.rb
```diff
@@ -96,6 +96,8 @@ def ruby_trunk_discourse(ssh, commit_hash, options)
   end
 
   def rails_releases(ssh, rails_version, options)
+    include_patterns = build_include_patterns(options[:include_patterns])
```
tgxworld: I just realized we don't need to assign it to a variable. Could we update this for the other jobs too? Thanks!

klgilbert: Do you want me to remove the build_include_patterns method and just call -e \"INCLUDE_PATTERNS=... directly

tgxworld: The method is called in a few other methods but there isn't a need to assign it to a variable.

Date: Wed, 17 Jun 2015 09:18:33 -0700

tgxworld:
Thanks!

PR #88, opened 2015-06-11
Fixed minor grammar errors

ttoti:
Fixed little grammatical errors to make it flow a little better ☺️

PR #128, opened 2015-11-04
Grammar fixes

tjschuck:
This is actually mostly a reversion of https://github.com/ruby-bench/ruby-bench-web/pull/88 -- not sure why that happened, but it's incorrect.  The benchmarks "are run", not "are ran".


tgxworld:
Thanks @tjschuck!

PR #136, opened 2016-02-03
Add favicon

yangshun:
@tgxworld Please patch and test locally. I don't have the dev env set up on this device so I just did an asset replacement >_<


tgxworld:
Thanks!!

PR #137, opened 2016-02-13
Changes in setup

sonalkr132:
Users will have different config for postgres (username, password etc).
It makes more sense to given them an example file which they use to create
their own config/database.yml. They won't have to worry about git tracking
the changes.

redis should start before we run any migration.
rake db:setup creates database too, there is no need of db:create.


config/database.yml.example
```diff
@@ -14,12 +14,15 @@
 # Configure Using Gemfile
 # gem 'pg'
 #
+# For details on connection pooling, see rails configuration guide
+# http://guides.rubyonrails.org/configuring.html#database-pooling
```
tgxworld: Any reason this needs to be moved up?

tgxworld:
@sonalkr132  Thanks for the PR :+1:  Just one minor comment.

config/database.yml.example
sonalkr132: That link doesn't really give much info about setting up pg database.

tgxworld: It provide details about connection pooling which I think should stay where it is. Plus it is the default anyway.

sonalkr132:
@tgxworld I have updated my PR

tgxworld:
@sonalkr132  Thanks!!! I'm sorting out the broken tests now.

PR #144, opened 2016-02-21
remove params again

lalithr95:
One more minor fix removed params which already calculated :smile: I like this project


lalithr95:
@tgxworld I see that accessing admin route requires http basic authentication. Where can I setup admin password ? I only see username as 'admin'

tgxworld:
:+1:

PR #170, opened 2016-06-06
added travis api script

shahsaurabh0605:
Added the script to trigger `.travis.yml` file in `ruby-bench-docker`.

### DIFF
--- a/app/services/travis_script.rb
+++ b/app/services/travis_script.rb
@@ -1,0 +1,15 @@
+require 'uri'
+require 'net/http'
+require 'json'
+
+uri = URI.parse("https://api.travis-ci.org/repo/ruby-bench%2Fruby-bench-docker/requests")
+http_client = Net::HTTP.new(uri.host, uri.port)
+http_client.use_ssl = true
+request = Net::HTTP::Post.new(uri.path)
+token = `travis token --org`.gsub(/\n/, "")
+request["Content-Type"] = 'application/json'
+request["Accept"] = 'application/json'
+request["Travis-API-Version"] = '3'
+request["Authorization"] = 'token '+token
+request.body = {'request' => {'branch' => 'master'}}.to_json
+response = http_client.request(request)


tgxworld:
Can't we do this in Ruby instead?

shahsaurabh0605:
Best way to do this should be in a script as indicated [here](https://docs.travis-ci.com/user/triggering-builds). For turning this to ruby we might need to use the `curb` which is the ruby wrapper for curl.

tgxworld:
That is just one way of doing it. Essentially, what you're doing is just a POST request which can be done in many ways. One such way is to use the built in library in Ruby http://ruby-doc.org/stdlib-2.3.1/libdoc/net/http/rdoc/Net/HTTP.html

shahsaurabh0605:
Made the changes in ruby. Have a look.

app/services/travis_script.rb
```diff
@@ -0,0 +1,14 @@
...
+require 'json'
+
+uri = URI.parse("https://api.travis-ci.org/repo/ruby-bench%2Fruby-bench-docker/requests")
+https = Net::HTTP.new(uri.host,uri.port)
+https.use_ssl = true
+request = Net::HTTP::Post.new(uri.path)
+request["Content-Type"] = 'application/json'
+request["Accept"] = 'application/json'
+request["Travis-API-Version"] = '3'
+request["Authorization"] = 'token smEOuqMAAUcwLHBnFjnJkA'
```
tgxworld: Does this token have to be a secret?

app/services/travis_script.rb
```diff
@@ -0,0 +1,14 @@
...
+request.body = {'request' => {'branch' => 'master'}}.to_json
+response = https.request(request)
```
tgxworld: What happens when the request returns a status other than `200`

app/services/travis_script.rb
```diff
@@ -0,0 +1,14 @@
+require 'uri'
+require 'net/http'
+require 'json'
+
+uri = URI.parse("https://api.travis-ci.org/repo/ruby-bench%2Fruby-bench-docker/requests")
```
tgxworld: What is with the encoding `%2` in the URL?

app/services/travis_script.rb
```diff
@@ -0,0 +1,14 @@
...
+https = Net::HTTP.new(uri.host,uri.port)
```
prathamesh-sonpatki: Nitpick but add a space between `uri.host` and `uri.port`.

prathamesh-sonpatki: Another nitpick, `https` doesn't sound right name. May be `http_client`?

prathamesh-sonpatki:
@shahsaurabh0605 Please also check http://chris.beams.io/posts/git-commit/. Writing good commit messages is very important so in future, people will know why you added this script :smile:

app/services/travis_script.rb
shahsaurabh0605: It's just the separator `%2F`. What of it?

app/services/travis_script.rb
shahsaurabh0605: I did not get this. Currently we get a success. Do I need to check for something?

app/services/travis_script.rb
tgxworld: Why do we need it `%2F` instead of just using`/` in string

app/services/travis_script.rb
tgxworld: We can not always assume success, if the URI is wrong or if the api is down. You'll get a different status code.

shahsaurabh0605: What should I do for the cases when there is no success?

app/services/travis_script.rb
shahsaurabh0605: It's just the same script I used from [here](https://docs.travis-ci.com/user/triggering-builds). So kept it that way.

PR #174, opened 2016-07-05
Detecting benchmark regressions

shahsaurabh0605:
Added a method to detect benchmark regressions. This fetches the result of last 10 commits from the database and compares with the newly created result.

### DIFF
--- a/app/controllers/benchmark_runs_controller.rb
+++ b/app/controllers/benchmark_runs_controller.rb
@@ -35,6 +35,10 @@ class BenchmarkRunsController < APIController
       $redis.del(key)
     end
 
+    if params[:commit_hash]
+      BenchmarkRegressionJob.new.perform(benchmark_run.id)
+    end
+
     render nothing: true
   end
 
--- a/app/jobs/benchmark_regression_job.rb
+++ b/app/jobs/benchmark_regression_job.rb
@@ -1,0 +1,75 @@
+require 'uri'
+require 'net/http'
+require 'json'
+
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(benchmark_run_id)
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+    categories = benchmark_run.result.keys
+    current_commit = benchmark_run.initiator
+    previous_commits = Commit.where('created_at < ?', current_commit.created_at).where(repo: current_commit.repo).limit(1000)
+    previous_benchmark_runs = BenchmarkRun.where(initiator: previous_commits, benchmark_type: benchmark_run.benchmark_type,
+                                                 benchmark_result_type: benchmark_run.benchmark_result_type)
+
+    categories.each do |category|
+      previous_benchmark_results = previous_benchmark_runs.map { |run| run.result[category].to_f }
+      results_average = average(previous_benchmark_results)
+      results_param = results_average + 2*standard_deviation(previous_benchmark_results, results_average)
+      if benchmark_run.result[category].to_f > results_param && check_equal(benchmark_run, category, results_param)
+        create_issue benchmark_run, category, results_param
+      end
+    end
+  end
+
+  def create_issue(benchmark_run, category, results_param)
+    uri = URI.parse(Rails.application.secrets.github_api)
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request.basic_auth 'RubyBenchBot', Rails.application.secrets.github_password
+    request.body = {"title" => "Benchmark Regression Detected",
+                    "body" => request_body(benchmark_run, category, results_param)}.to_json
+    response = http_client.request(request)
+    raise StandardError, "Request failed with response code #{response.code}" if response.code != "201"
+    response
+  end
+
+  def average(previous_benchmark_results)
+    previous_benchmark_results.inject{|sum, el| sum + el}.to_f / previous_benchmark_results.size
+  end
+
+  def standard_deviation(previous_benchmark_results, results_average)
+    Math.sqrt(previous_benchmark_results.inject(0){|accum, i| accum + (i - results_average)**2} / (previous_benchmark_results.size - 1).to_f)
+  end
+
+  def request_body(benchmark_run, category, results_param)
+    "Benchmark Regression has been detected:
+    repo: #{benchmark_run.initiator.repo.name}
+    benchmark_category: #{category}
+    benchmark_type: #{benchmark_run.benchmark_type.category}
+    benchmark_result_type: #{benchmark_run.benchmark_result_type.name}
+    Diversion: #{((benchmark_run.result[category].to_f - results_param) / results_param) * 100}%"
+  end
+
+  def check_equal(benchmark_run, category, results_param)
+    url = URI.parse(Rails.application.secrets.github_api+"?state=open&since=#{get_time}")
+    http = Net::HTTP.new(url.host, url.port)
+    request = Net::HTTP::Get.new(url.to_s)
+    http.use_ssl = true
+    res = http.request(request)
+    raise StandardError, "Request failed with response code #{res.code}" if res.code != "200"
+    response = JSON.parse(res.body)
+
+    response.each do |response|
+      if request_body(benchmark_run, category, results_param).split(/Diversion/).first.eql? response["body"].split(/Diversion/).first
+        return false
+      end
+    end
+  end
+
+  def get_time
+    time_before_iso = Time.parse((Time.current - (60 * 60 * 24 * 30)).to_s).utc.iso8601
+  end
+end
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -15,6 +15,8 @@ shared: &shared
   bare_metal_server_user: <%= ENV['BARE_METAL_SERVER_USER'] %>
   bare_metal_server_password: <%= ENV['BARE_METAL_SERVER_PASSWORD'] %>
   heroku_api_key: <%= ENV['HEROKU_API_KEY'] %>
+  github_api: <%= ENV["GITHUB_API"] %>
+  github_password: <%= ENV["GITHUB_PASSWORD"] %>
 
 development:
   <<: *shared
--- a/test/jobs/benchmark_regression_job_test.rb
+++ b/test/jobs/benchmark_regression_job_test.rb
@@ -1,0 +1,40 @@
+require 'test_helper'
+
+class BenchmarkRegressionJobTest < ActiveJob::TestCase
+  setup do
+    @array = [323, 11, 555, 666, 234, 21, 666, 343, 1, 2]
+    benchmark_result_type = create(:benchmark_result_type)
+    benchmark_type = create(:benchmark_type)
+    repo = create(:repo)
+    10.times do
+      commit = create(:commit, repo: repo)
+      @benchmark_run = create(:commit_benchmark_run,
+                              benchmark_result_type: benchmark_result_type,
+                              benchmark_type: benchmark_type,
+                              initiator: commit)
+    end
+  end
+
+  test "create issue for benchmark regression" do
+    VCR.use_cassette('benchmark_regression') do
+      response = BenchmarkRegressionJob.new.create_issue(@benchmark_run, @benchmark_run.result.keys[0], 4.0)
+      assert_equal "201", response.code
+    end
+  end
+
+  test "calculates average" do
+    assert_equal 282.2, BenchmarkRegressionJob.new.average(@array)
+  end
+
+  test "calculates standard_deviation" do
+    assert_equal 274.08, BenchmarkRegressionJob.new.standard_deviation(@array, BenchmarkRegressionJob.new.average(@array)).round(2)
+  end
+
+  test "check for similar issues" do
+    travel_to "2016-07-03 13:50:41 +0530" do
+      VCR.use_cassette('similar_issues') do
+        assert BenchmarkRegressionJob.new.check_equal(@benchmark_run, @benchmark_run.result.keys[0], 4.0)
+      end
+    end
+  end
+end
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,7 @@
 /log/*.log
 /tmp
 
-/test/vcr_cassettes/**/*.yml
+#/test/vcr_cassettes/**/*.yml
 *.rdb
 
 # Ignore user specific database config


tgxworld:
Hmm have you seen my post in http://community.rubybench.org/t/gsoc-project-improving-rubybench/99/53?u=tgxworld? :smile:

app/controllers/benchmark_runs_controller.rb
```diff
@@ -27,10 +27,22 @@ def create
...
     )
 
+    if params[:commit_hash]
+      commit_objects = BenchmarkRun.where( initiator: initiator, benchmark_type: benchmark_type,
+                       benchmark_result_type: benchmark_result_type).limit(10)
+      commit_results = Array.new
+      commit_objects.each do |object|
+        commit_results << object.result[category]
+      end
+    end
```
tgxworld: All this should be done in a background job

app/controllers/benchmark_runs_controller.rb
```diff
@@ -27,10 +27,22 @@ def create
...
+
     benchmark_run.update_attributes(benchmark_run_params)
     benchmark_run.result = params[:benchmark_run][:result]
     benchmark_run.save!
-
+    
+    if benchmark_run.result[category] > commit_results.min
+      # create an issue in ruby-bench-suite
+    end
```
prathamesh-sonpatki: This should also go to background job.

app/controllers/benchmark_runs_controller.rb
prathamesh-sonpatki: The background job can be called after the current `benchmark_run` is saved to the database successfully.

app/controllers/benchmark_runs_controller.rb
```diff
@@ -27,10 +27,22 @@ def create
       benchmark_result_type: benchmark_result_type
     )
 
+    if params[:commit_hash]
+      commit_objects = BenchmarkRun.where( initiator: initiator, benchmark_type: benchmark_type,
+                       benchmark_result_type: benchmark_result_type).limit(10)
```
prathamesh-sonpatki: The number 10 should not be hard coded. I know it's still WIP but just noting it down :smile:

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,16 @@
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+ 
+  def perform(initiator, benchmark_type, benchmark_result_type, category)
+    commit_objects = BenchmarkRun.where( initiator: initiator, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).limit(10)
```
shahsaurabh0605: What should be the best value here instead of 10?

tgxworld: What values have you considered and why?

shahsaurabh0605: I think we can hardcode it to a specific number but this largely depends on what baseline we are considering to detect performance regression in this new commit.

tgxworld: What method have you decided to implement to detect regressions? Based on the method, you can them derive a reasonable value.

shahsaurabh0605: I added a post in http://community.rubybench.org/t/gsoc-project-improving-rubybench/99/55 to discuss the method.

shahsaurabh0605:
Have a look at this.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,20 @@
...
+    commit_results = Array.new
+    commit_objects.drop(1).each do |object|
+      commit_results << object.result[category]
+    end
+    
+    results_avg = commit_results.inject{ |sum, el| sum + el }.to_f / commit_results.size
+    results_standard_deviation = Math.sqrt(commit_results.inject(0){|accum, i| accum +(i-results_avg)**2 }/(commit_results.length - 1).to_f)
+    results_param = results_avg + 2*results_standard_deviation
+
+    if commit_objects[0].result[category] > results_param
```
tgxworld: I don't think the new result should be included in the calculation of the standard deviation

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,20 @@
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+ 
+  def perform(initiator, benchmark_type, benchmark_result_type, category)
+    commit_objects = BenchmarkRun.where( initiator: initiator, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).limit(1000)
```
tgxworld: If another benchmark result is submitted before this job has been executed, the result will be inaccurate.

app/jobs/benchmark_regression_job.rb
shahsaurabh0605: I am not including the new result in the calculation of standard deviation by dropping the first element in the loop (see `commit_objects.drop(1)`)

tgxworld: ahh ic. Anyway, the default scope is to order by `created_at`, what we want here is to order by `Commit#created_at` instead.

app/jobs/benchmark_regression_job.rb
shahsaurabh0605: Can another benchmark result be submitted so fast? Because the only race condition will be while fetching the data which itself is quite fast.

tgxworld: Right now we only have one server that is running the benchmarks, benchmark results will come in orderly. However, if this job fails for whatever reason and we need to retry it. The benchmark_runs that you fetch might be a totally different set of data.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,23 @@
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+ 
+  def perform(initiator, benchmark_type, benchmark_result_type, category)
+    commit_objects = BenchmarkRun.where(benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).offset(10).limit(1000)
```
shahsaurabh0605: Since, we are fetching 1000 results it doesn't make much of a difference if we leave a first few results. So we can set a safety offset(here 10) so that we can be sure that we are fetching the correct results.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,23 @@
...
+    commit_results = Array.new
+    commit_objects.drop(1).each do |object|
+      commit_results << object.result[category]
+    end
+    
+    results_avg = commit_results.inject{ |sum, el| sum + el }.to_f / commit_results.size
+    results_standard_deviation = Math.sqrt(commit_results.inject(0){|accum, i| accum +(i-results_avg)**2 }/(commit_results.length - 1).to_f)
+    results_param = results_avg + 2*results_standard_deviation
+
+    current = BenchmarkRun.find_by(initiator: initiator)
```
shahsaurabh0605: We can find the benchmark for which the job is running this way, right?

app/jobs/benchmark_regression_job.rb
tgxworld: > Since, we are fetching 1000 results it doesn't make much of a difference if we leave a first few results.

It does make a difference if the last 10 results are way off. What you want here is the benchmark results of  the last 1000 commits which you can easily fetch.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,23 @@
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+ 
+  def perform(initiator, benchmark_type, benchmark_result_type, category)
+    commit_objects = BenchmarkRun.where(benchmark_type: benchmark_type,
```
tgxworld: You are fetching benchmark runs here but the variable is named `commit_objects` which is pretty confusing

app/jobs/benchmark_regression_job.rb
tgxworld: Hmm I don't quite understand what you mean.

`BenchmarkRun.find_by(initiator: initiator)` will fetch all benchmark runs that belongs to a particular commit

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    current_commit = BenchmarkRun.find_by(initiator: initiator, initiator_type: Commit, benchmark_type: benchmark_type, benchmark_result_type: benchmark_result_type)       
+
+    current_initiator = current_commit.initiator_id
+
+    last_initiator = BenchmarkRun.where(initiator_type: Commit, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).order(:created_at).first.initiator_id
+
+    benchmark_runs = BenchmarkRun.where(initiator_type: Commit, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).offset(last_initiator - current_initiator).limit(1000)
```
shahsaurabh0605: This sets the offset so that the benchmarks which came after our benchmark are not considered.

app/controllers/benchmark_runs_controller.rb
```diff
@@ -30,6 +30,10 @@ def create
     benchmark_run.update_attributes(benchmark_run_params)
     benchmark_run.result = params[:benchmark_run][:result]
     benchmark_run.save!
+ 
+    if params[:commit_hash]
+      BenchmarkRegressionJob.perform_later(initiator, benchmark_type, benchmark_result_type, category)
```
tgxworld: I believe you can pass in the benchmark_run created here and it'll make things alot easier

shahsaurabh0605:
Done :+1:

tgxworld:
Is this PR done?

shahsaurabh0605:
You can have a look at it once. According to me, it's done :smile:

tgxworld:
hmm I see WIP in the title and we don't seem to be creating an issue in Github yet when a regression is detected? You need to add tests as well otherwise, we won't know that your code is working as intended.

shahsaurabh0605:
Done :+1:

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,17 @@
...
+    initiator = commit
+    benchmark_run = create(:commit_benchmark_run,
+      benchmark_result_type: benchmark_result_type,
+      benchmark_type: benchmark_type,
+      initiator: commit
+    )
+    category = "sometime"
+    BenchmarkRegressionJob.new.perform(benchmark_run, initiator, benchmark_type, benchmark_result_type, category)
+  end
+end
```
tgxworld: erm.... what exactly is this test testing for?

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,26 @@
...
+    previous_benchmark_runs.drop(1).each do |run|
+      commit_results << run.result[category].to_f
+    end
+
+    results_avg = commit_results.inject{ |sum, el| sum + el }.to_f / commit_results.size
+    results_standard_deviation = Math.sqrt(commit_results.inject(0){|accum, i| accum +(i-results_avg)**2 }/(commit_results.length - 1).to_f)
+    results_param = results_avg + 2*results_standard_deviation
+
+    if benchmark_run.result[category].to_f > results_param
+      `curl --user "RubyBenchBot:a12345a" -X POST --data '{"title":"Benchmark Regression Detected"}' https://api.github.com/repos/ruby-bench/ruby-bench-suite/issues`
```
tgxworld: Please look up how HTTP calls are usually made in Ruby. We usually don't do it via a system call to `curl`

test/jobs/benchmark_regression_job_test.rb
shahsaurabh0605: It's testing for the new benchmark regression job which I added!

tgxworld: Can you please read through http://guides.rubyonrails.org/testing.html.

I am not following how your test is actually testing that the job you added works

app/jobs/benchmark_regression_job.rb
prathamesh-sonpatki: Also better to move this in separate class.

test/jobs/benchmark_regression_job_test.rb
shahsaurabh0605: How do I add tests for this job differently? Currently this test just runs the perform method with appropriate parameters to test the method which is similar to what implemented in remote_server_job tests.

tgxworld: It isn't similar to `remote_server_job` tests. Note how those tests sets expectation about certain things.

tgxworld: What you're doing here is just calling the method. 

* How do you know if it is correctly detecting a regression?
* When it correctly detects a regression, what is the code expected to do?

These have to be included in the tests

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,41 @@
...
+    results_standard_deviation = Math.sqrt(commit_results.inject(0){|accum, i| accum +(i-results_avg)**2 }/(commit_results.length - 1).to_f)
+    results_param = results_avg + 2*results_standard_deviation
+
+    if benchmark_run.result[category].to_f > results_param
+      create_issue
+    end
+  end
+
+  def create_issue
+    uri = URI.parse("https://api.github.com/repos/shahsaurabh0605/docker-practice/issues")
```
shahsaurabh0605: For the tests, issue will be opened once for the first time and then tests will go normal without making a post request.

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,43 @@
+require 'test_helper'
+
+class BenchmarkRegressionJobTest < ActiveJob::TestCase
+
+  test "No Benchmark Regression Detected" do
+    benchmark_run, initiator, benchmark_type, benchmark_result_type = setup_benchmarks
```
tgxworld: Check out https://www.relishapp.com/rspec/rspec-core/v/2-2/docs/hooks/before-and-after-hooks

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,43 @@
...
+      assert_kind_of(Net::HTTPSuccess, BenchmarkRegressionJob.new.perform(benchmark_run, initiator, benchmark_type, benchmark_result_type))
+    end
+  end
+
+  private
+
+  def setup_benchmarks
+    benchmark_result_type = create(:benchmark_result_type)
+    benchmark_type = create(:benchmark_type)
+    (1...1000).each do
```
tgxworld: We don't really need 1000 records here. This will make the tests extremely slow. 10 will do.

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,43 @@
...
+        initiator: commit
+      )
+    end
+    commit = create(:commit)
+    initiator = commit
+    benchmark_run = create(:commit_benchmark_run,
+      benchmark_result_type: benchmark_result_type,
+      benchmark_type: benchmark_type,
+      initiator: commit
+    )
```
tgxworld: Is line 34 - 40 needed?

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,43 @@
...
+    benchmark_run.result = {"sometime"=>"5.1"}
+    VCR.use_cassette('benchmark_regression') do
+      assert_kind_of(Net::HTTPSuccess, BenchmarkRegressionJob.new.perform(benchmark_run, initiator, benchmark_type, benchmark_result_type))
+    end
+  end
+
+  private
+
+  def setup_benchmarks
+    benchmark_result_type = create(:benchmark_result_type)
```
tgxworld: In Rspec, we prefer to use [let](https://www.relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let)

app/controllers/benchmark_runs_controller.rb
```diff
@@ -31,6 +31,10 @@ def create
     benchmark_run.result = params[:benchmark_run][:result]
     benchmark_run.save!
 
+    if params[:commit_hash]
+      BenchmarkRegressionJob.perform_later(benchmark_run, initiator, benchmark_type, benchmark_result_type, benchmark_type_params[:category])
```
tgxworld: Please read this and see what can be improved here.

https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,41 @@
...
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request.basic_auth 'RubyBenchBot', 'a12345a'
+    request.body = {'title' => 'Benchmark Regression Detected'}.to_json
+    response = http_client.request(request)
```
tgxworld: What happens if the response status is not `200`?

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,41 @@
...
+      create_issue
+    end
+  end
+
+  def create_issue
+    uri = URI.parse("https://api.github.com/repos/shahsaurabh0605/docker-practice/issues")
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request.basic_auth 'RubyBenchBot', 'a12345a'
```
tgxworld: You have to change your password, that should have been passed in as a secret

test/jobs/benchmark_regression_job_test.rb
tgxworld: You can call this in a [setup method](https://github.com/seattlerb/minitest#unit-tests)

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,43 @@
...
+    VCR.use_cassette('benchmark_regression') do
+      assert_nil BenchmarkRegressionJob.new.perform(benchmark_run, initiator, benchmark_type, benchmark_result_type)
+    end
+  end
+
+  test "Benchmark Regression Detected" do
+    benchmark_run, initiator, benchmark_type, benchmark_result_type = setup_benchmarks
+    benchmark_run.result = {"sometime"=>"5.1"}
+    VCR.use_cassette('benchmark_regression') do
+      assert_kind_of(Net::HTTPSuccess, BenchmarkRegressionJob.new.perform(benchmark_run, initiator, benchmark_type, benchmark_result_type))
```
tgxworld: You want to test for what is in the response and not just the `class` of it

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,41 @@
...
+    category = benchmark_run.result.keys[0]
+    current_initiator = benchmark_run.initiator_id
+    last_initiator = BenchmarkRun.where(initiator_type: Commit, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).order(:created_at).first.initiator_id
+
+    previous_benchmark_runs = BenchmarkRun.where(initiator_type: Commit, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).offset(last_initiator - current_initiator).limit(1000)
+
+    commit_results = Array.new
+    previous_benchmark_runs.drop(1).each do |run|
```
tgxworld: you can do `previous_benchmark_runs.drop(1).map` instead

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def perform(benchmark_run, initiator, benchmark_type, benchmark_result_type)
+
+    category = benchmark_run.result.keys[0]
+    current_initiator = benchmark_run.initiator_id
+    last_initiator = BenchmarkRun.where(initiator_type: Commit, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).order(:created_at).first.initiator_id
+
+    previous_benchmark_runs = BenchmarkRun.where(initiator_type: Commit, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).offset(last_initiator - current_initiator).limit(1000)
```
tgxworld: > `.offset(last_initiator - current_initiator)` 

This is incorrect. You should use the created_at date instead.

app/controllers/benchmark_runs_controller.rb
shahsaurabh0605: I have reduced to only 3 parameters `benchmark_run, benchmark_type, benchmark_result_type`. Doing something as shown in the link will be difficult for us as all this parameters are obtained as a result of many different methods.

tgxworld: It is not how many parameters that is the important bit. It is what you pass in :wink:

app/controllers/benchmark_runs_controller.rb
```diff
@@ -31,6 +31,12 @@ def create
     benchmark_run.result = params[:benchmark_run][:result]
     benchmark_run.save!
 
+    if params[:commit_hash]
+      BenchmarkRegressionJob.new.perform(commit_hash: params[:commit_hash], category: benchmark_type_params[:category],
+      script_url: benchmark_type_params[:script_url], name: benchmark_result_type_params[:name],
+      unit: benchmark_result_type_params[:unit], repos: params[:repo], organizations: params[:organization] )
+    end
```
shahsaurabh0605: Can this be more optimized? Also, for writing the tests I need to change the benchmark results and check. But with this parameters how can I write tests?

app/jobs/benchmark_regression_job.rb
shahsaurabh0605: What should I do if response status is not success? Should I just leave a message or try again?

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def perform(params = {})
+
+    repo = Repo.joins(:organization)
+      .where(name: params[:repos], organizations: { name: params[:organizations] })
+      .first
+
+    initiator = repo.commits.find_by_sha1(params[:commit_hash])
+
+    benchmark_type = repo.benchmark_types.find_or_create_by!(
```
tgxworld: Why are we creating the record if it can't be found here?

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,59 @@
...
+      category: params[:category],
+      script_url: params[:script_url]
+    )
+  
+    benchmark_result_type = BenchmarkResultType.find_or_create_by!(
+      name: params[:name],
+      unit: params[:unit]
+    )
+  
+    benchmark_run = BenchmarkRun.find_or_initialize_by(
```
tgxworld: Why are we initializing the record if it can't be found here?

app/controllers/benchmark_runs_controller.rb
tgxworld: You already have the `benchmark_run` and the other records. Just pass in their ID and do a `find(id)` in the job. 

Not sure if that wasn't clear in https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def create_issue
+    uri = URI.parse("https://api.github.com/repos/shahsaurabh0605/docker-practice/issues")
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request.basic_auth 'RubyBenchBot', Rails.application.secrets.github_password
+    request.body = {'title' => 'Benchmark Regression Detected'}.to_json
+    http_client.request(request)
```
tgxworld: If the response is not 200, raise an error

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,44 @@
...
+    results_standard_deviation = Math.sqrt(commit_results.inject(0){|accum, i| accum +(i-results_avg)**2 }/(commit_results.length - 1).to_f)
+    results_param = results_avg + 2*results_standard_deviation
+
+    if benchmark_run.result[category].to_f > results_param
+      create_issue
+    end
+  end
+
+  def create_issue
+    uri = URI.parse("https://api.github.com/repos/shahsaurabh0605/docker-practice/issues")
```
shahsaurabh0605: Have a look. If everything seems fine, I will change this.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,44 @@
...
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request.basic_auth 'RubyBenchBot', Rails.application.secrets.github_password
+    request.body = {'title' => 'Benchmark Regression Detected'}.to_json
+    response = http_client.request(request)
+    raise StandardError, "Request failed with response code #{response.code}" if response.code != "201"
```
tgxworld: Why 201?

config/secrets.yml
```diff
@@ -36,5 +37,6 @@ production:
   secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
   api_name: <%= ENV["API_NAME"] %>
   api_password: <%= ENV["API_PASSWORD"] %>
+  github_password: <%= ENV["GITHUB_PASSWORD"] %>
```
shahsaurabh0605: I have set my github password in bashrc file and using it as an environment variable here. This is the way to keep our passwords secret, right?

app/jobs/benchmark_regression_job.rb
shahsaurabh0605: It is `creating` an issue with response code 201

config/secrets.yml
tgxworld: We should only make the request in production.

shahsaurabh0605: How will the tests pass for travis ci?

app/jobs/benchmark_regression_job.rb
tgxworld: Icic had to check the API to realize that

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,44 @@
...
+  queue_as :default
+
+  def perform(benchmark_run_id, benchmark_type_id, benchmark_result_type_id)
+
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+    benchmark_type = BenchmarkType.find_by_id(benchmark_type_id)
+    benchmark_result_type = BenchmarkResultType.find_by_id(benchmark_result_type_id)
+
+    category = benchmark_run.result.keys[0]
+    current_id = benchmark_run_id
```
tgxworld: No need for this variable. It is duplicating stuff

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,44 @@
...
+    last_id = BenchmarkRun.where(initiator_type: Commit, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).order(:created_at).first.id
+
+    previous_benchmark_runs = BenchmarkRun.where("id not in (?)",(current_id..last_id)).where(initiator_type: Commit, benchmark_type: benchmark_type,
```
tgxworld: I mentioned this previously. You can't use the id range to fetch the benchmark runs. You need to fetch the benchmark runs of the last 1000 commits ordered by the created at date of the commit.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,44 @@
...
+require 'net/http'
+require 'json'
+
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(benchmark_run_id, benchmark_type_id, benchmark_result_type_id)
+
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+    benchmark_type = BenchmarkType.find_by_id(benchmark_type_id)
```
tgxworld: `benchmark_run` belongs to `benchmark_type` and `benchmark_result_type`. I don't think you need to be fetching the records with the ids

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,44 @@
...
+                     benchmark_result_type: benchmark_result_type).order(:created_at).first.id
+
+    previous_benchmark_runs = BenchmarkRun.where("id not in (?)",(current_id..last_id)).where(initiator_type: Commit, benchmark_type: benchmark_type,
+                     benchmark_result_type: benchmark_result_type).limit(1000)
+
+    commit_results = previous_benchmark_runs.map { |run| run.result[category].to_f }
+
+    results_avg = commit_results.inject{ |sum, el| sum + el }.to_f / commit_results.size
+    results_standard_deviation = Math.sqrt(commit_results.inject(0){|accum, i| accum +(i-results_avg)**2 }/(commit_results.length - 1).to_f)
+    results_param = results_avg + 2*results_standard_deviation
```
tgxworld: Please extract the logic here into a different class and add unit tests for it

app/jobs/benchmark_regression_job.rb
shahsaurabh0605: Previously, I was doing with benchmark_ids, which I realized was incorrect. Now, I just remove the ids which came after my benchmark_run so that if another benchmark runs before this job is executed it won't be a problem. How do I achieve this by ordered by created at?

config/secrets.yml
tgxworld: You need to check in the VCR response.

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,39 @@
...
+    @benchmark_run.update(result: {"sometime"=>"5.1"})
+    VCR.use_cassette('benchmak_regression') do
+      response = BenchmarkRegressionJob.new.perform(@benchmark_run.id, @benchmark_type.id, @benchmark_result_type.id )
+      assert_equal "201" , response.code
+    end
+  end
+
+  private
+
+  def setup_benchmarks
```
tgxworld: This whole block can be moved into the setup block

shahsaurabh0605:
Made the changes.

tgxworld:
@shahsaurabh0605  Specs are failing.

shahsaurabh0605:
Specs are failing because whenever it tries to create a issue, password in secrets.yml for test environment isn't correct. How can I solve this?

prathamesh-sonpatki:
@shahsaurabh0605 We can use [VCR](https://github.com/vcr/vcr) as @tgxworld suggested. You need to check in the VCR file generated locally in the git so that when tests are run next time, the request will not be made to Github API.

app/jobs/benchmark_regression_job.rb
prathamesh-sonpatki: Can this also be moved to a ENV var or a constant?

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,46 @@
...
+    if benchmark_run.result[category].to_f > results_param
+      create_issue
+    end
+  end
+
+  def create_issue
+    uri = URI.parse("https://api.github.com/repos/shahsaurabh0605/docker-practice/issues")
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
```
prathamesh-sonpatki: @shahsaurabh0605 Did you explore other HTTP clients?

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,46 @@
...
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(benchmark_run_id)
+
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+
+    category = benchmark_run.result.keys[0]
+
+    previous_benchmark_runs = BenchmarkRun.where('created_at < ?', benchmark_run.created_at).where(initiator_type: Commit, benchmark_type: benchmark_run.benchmark_type,
```
tgxworld: > You can't use the id range to fetch the benchmark runs. You need to fetch the benchmark runs of the last 1000 commits ordered by the created at date of the commit.

I'll try to make it clearer.... 

1. Find the lastest 1000 commits **of the right repository** for the given benchmark_type and benchmark_result_type
2. Get the benchmark_run of those commits and form the result.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  def create_issue
+    uri = URI.parse("https://api.github.com/repos/shahsaurabh0605/docker-practice/issues")
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request.basic_auth 'RubyBenchBot', Rails.application.secrets.github_password
+    request.body = {'title' => 'Benchmark Regression Detected'}.to_json
```
tgxworld: You need to include more details in the issue in order for the dev to quickly identify which benchmark has regressed

app/jobs/benchmark_regression_job.rb
shahsaurabh0605: Looking at the model of benchmark_runs above query should fetch the required results. As we are giving benchmark type as a parameter why do we need to check for the right repository?

tgxworld: > Looking at the model of benchmark_runs above query should fetch the required results.

Can you explain in your own words what we are trying to query for here? I still don't you understand what we need to be fetching here :crying_cat_face: 

> As we are giving benchmark type as a parameter why do we need to check for the right repository?
I didn't say anything about checking :disappointed: The emphasis is to let you know that you have to ensure that the benchmark_runs are scope to the right repository.

shahsaurabh0605: screenshot from 2016-07-22 11-33-35

Say, for this database of benchmark_runs, we need all the results for benchmark category `ao_bench` then we need to give initiator type, benchmark_type and benchmark_result_type to fetch our required result of `"0.22"`. This way we get all the 1000 results. Giving these queries we are already in **the right repository**.

tgxworld: Ok I see where your confusion is. Benchmark Runs are not always created in the right order. For example, our server can be running a benchmark_run for commit 2 before it runs the benchmark for commit 1. Therefore, I've reiterated a couple of times that you need to `Find the latest 1000 commits` for that particular benchmark and fetch the benchmark_runs from there.

shahsaurabh0605: That cleared a bit. Another query that I provided was to fetch only the required benchmark runs which are created before our `commit` of interest. This ensures that all the commits which are created after our `commit` and before our benchmark regression job is finished are handled and not taken into account. So for a server running commit 2 first instead of commit 1 will have a benchmark regression test for commit 2 first and then commit 1. Is there a problem?

tgxworld: Ok so what you need to do instead is to fetch last 1000 commits sorted by the `commits.created_at` and `commits.created_at <= <the current commit date>`

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def create_issue(benchmark_run)
+    uri = URI.parse(Rails.application.secrets.github_api)
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request.basic_auth 'RubyBenchBot', Rails.application.secrets.github_password
+    request.body = {"title" => "Benchmark Regression Detected",
+    "body" => "This benchmark regression is caused due to commit #{benchmark_run.initiator.sha1} from the repository #{benchmark_run.initiator.repo.name}"}.to_json
```
tgxworld: This still doesn't tell us what benchmark has regressed.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    results_param = average(previous_benchmark_results) + 2*standard_deviation(previous_benchmark_results)
+
+    if benchmark_run.result[category].to_f > results_param
+      create_issue benchmark_run
+    end
+  end
+
+  def create_issue(benchmark_run)
+    uri = URI.parse(Rails.application.secrets.github_api)
```
tgxworld: Why does GitHub's API endpoint need to be in a secret?

shahsaurabh0605: @prathamesh-sonpatki suggested to use this as an environment variable too.

prathamesh-sonpatki: Yes I suggested it because, I saw the earlier URL was pointing to Saurabh's account and there was a comment about changing it later to the actual Ruby Bench repo. 

By having ENV var, Saurabh can test locally on his repo but production repo will be Ruby Bench repo. Also it will be useful for us for testing as well :smile:

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,50 @@
...
+require 'json'
+
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(benchmark_run_id)
+
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+
+    category = benchmark_run.result.keys[0]
```
tgxworld: This is incorrect. When the result has more than one key, we'll not be checking for it

shahsaurabh0605: Why will the result have more than one key? For a particular benchmark run undergoing this job there is only one key possible.

tgxworld: It isn't clear because the result is stored in JSON

https://github.com/ruby-bench/ruby-bench-web/blob/master/app/services/chart_builder.rb#L15-L19

shahsaurabh0605: So let's say if there are two categories. Then do we need to check for both the categories and create an issue if any one of them fails? Basically checking in a loop for all the categories.

tgxworld: Yes that is correct.

shahsaurabh0605:
Have a look at the changes. I can't check if travis will pass with the vcr cassettes because of build error.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,52 @@
...
+  def perform(benchmark_run_id)
+
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+
+    category = benchmark_run.result.keys.map {|key| key}
+
+    current_commit = benchmark_run.initiator
+    previous_commits = Commit.where('created_at < ?', current_commit.created_at).where(repo: current_commit.repo).limit(1000)
+
+    previous_benchmark_runs = previous_commits.map {|commit| BenchmarkRun.find_by(initiator: commit, benchmark_type: benchmark_run.benchmark_type,
```
tgxworld: We usually use a `do ... end` for multiline blocks

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,52 @@
...
+require 'json'
+
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(benchmark_run_id)
+
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+
+    category = benchmark_run.result.keys.map {|key| key}
```
tgxworld: `categories` would be a better variable name.

`benchmark_run.result.keys.map {|key| key}` is equivalent to `benchmark_run.result.keys`

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def create_issue(benchmark_run)
+    uri = URI.parse(Rails.application.secrets.github_api)
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request.basic_auth 'RubyBenchBot', Rails.application.secrets.github_password
+    request.body = {"title" => "Benchmark Regression Detected",
+    "body" => "This benchmark regression is caused due to commit #{benchmark_run.initiator.sha1} from the repository #{benchmark_run.initiator.repo.name} for the benchmark #{benchmark_run.result.keys[0]}"}.to_json
```
tgxworld: `benchmark_run.result.keys[0]` is incorrect here

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  test "calculates standard_deviation" do
+    assert_equal 274.08, BenchmarkRegressionJob.new.standard_deviation(@array).round(2)
+  end
+
+  test "#perform benchmark regression job" do
+    assert_equal @benchmark_run.result.keys , BenchmarkRegressionJob.new.perform(@benchmark_run.id)
+  end
+
+  test "create issue for benchmark regression" do
+    VCR.use_cassette('benChmark_regression') do
```
tgxworld: Typo in `benChmark_regression`

tgxworld:
I've fixed the build. Can you rebase?

test/jobs/benchmark_regression_job_test.rb
shahsaurabh0605: That's not a typo. Since it's a temporary github api, I am using this cassette for now.

shahsaurabh0605:
It seems that travis cannot fetch the URI which is stored as environment variable.

tgxworld:
Is your test passing locally? The error is `URI::InvalidURIError: bad URI(is not URI?)` which means the URI you are giving travis is incorrect

shahsaurabh0605:
Yes the tests are passing locally because the uri exists in my bashrc file as an environement variable, so `Rails.application.secrets.github_api` is able to fetch it.

test/jobs/benchmark_regression_job_test.rb
tgxworld: The capitalize `C` is not a typo?

tgxworld:
Ok so can you figure out how to pass the ENV variable to travis?

shahsaurabh0605:
I have commented env variable settings in `.travis.yml` file. We just need to set the github api, encrypt github password and add the secure key there. But the secure key can't be added by pull request from forks, so you need to add it there.

test/jobs/benchmark_regression_job_test.rb
shahsaurabh0605: It's just any random cassette name given by me for testing. I will change it to `Benchmark_Regression` once we set the correct api. :smile:

tgxworld:
I am not following. Why will we need to set the env variable in travis if we're using the cassettee?

shahsaurabh0605:
According to me, this is what is happening. Say, if I hardcode github_api or github_password in my test environment then `Rails.application.secrets.github_api` will fetch the api and recorded cassette will play. But what if I change the api in test environment? Then the cassette won't play because the URI got changed. 

Same thing is happening here. For local tests I set the env variables and it works fine. Cassettes don't create an issue again and again and just recording plays. But since the env variables are not known in travis, we get bad URI error because the api doesn't exist. VCR cassettes play when the github api remains the same and useful for not creating an issue again and again.

tgxworld:
let's hardcore `https://github.com/ruby-bench/github-api-sandbox` for test environment

shahsaurabh0605:
What about the password? I can't hardcore the password, right?

tgxworld:
Let me know exactly what I need to do? :)

shahsaurabh0605:
If we hardcore the api for the test environment then we will have to hardcore password as well. And we can't hardcore the password even for test environment. So we can't do this.

tgxworld:
Ok so do you need me to create a GitHub account to act as the bot? with user and password?

shahsaurabh0605:
Yes... I have created it, can we work with that?

tgxworld:
Sure PM me on community.rubybench.org and I'll add it to the env variables

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  test "calculates standard_deviation" do
+    assert_equal 274.08, BenchmarkRegressionJob.new.standard_deviation(@array).round(2)
+  end
+
+  test "#perform benchmark regression job" do
+    assert_equal @benchmark_run.result.keys , BenchmarkRegressionJob.new.perform(@benchmark_run.id)
+  end
+
+  test "create issue for benchmark regression" do
+    VCR.use_cassette('Benchmark_Regression') do
```
tgxworld: Don't capitalize the name of the cassettee

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  test "calculates average" do
+    assert_equal 282.2, BenchmarkRegressionJob.new.average(@array)
+  end
+
+  test "calculates standard_deviation" do
+    assert_equal 274.08, BenchmarkRegressionJob.new.standard_deviation(@array).round(2)
+  end
+
+  test "#perform benchmark regression job" do
+    assert_equal @benchmark_run.result.keys , BenchmarkRegressionJob.new.perform(@benchmark_run.id)
```
tgxworld: Nuke this test... It isn't testing for anything meaningful

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,57 @@
...
+  def create_issue(benchmark_run, category)
+    puts "ENV: #{ENV['GITHUB_API']}"
+    puts ENV.to_h
+    uri = URI.parse(Rails.application.secrets.github_api)
+    http_client = Net::HTTP.new(uri.host, uri.port)
+    http_client.use_ssl = true
+    request = Net::HTTP::Post.new(uri.path)
+    request.basic_auth 'RubyBenchBot', Rails.application.secrets.github_password
+    request.body = {"title" => "Benchmark Regression Detected",
+    "body" => "This benchmark regression is caused due to commit #{benchmark_run.initiator.sha1} from the repository #{benchmark_run.initiator.repo.name} for the benchmark #{category}"}.to_json
```
shahsaurabh0605: Is this format for the body of issue done or any changes to be made?

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,57 @@
...
+                       benchmark_result_type: benchmark_run.benchmark_result_type)
+    end
+
+    categories.each do |category|
+      previous_benchmark_results = previous_benchmark_runs.map { |run| run.result[category].to_f }
+
+      results_param = average(previous_benchmark_results) + 2*standard_deviation(previous_benchmark_results)
+
+      if benchmark_run.result[category].to_f > results_param
+        create_issue benchmark_run, category
```
tgxworld: I don't want to create multiple issue per `benchmark_run`. The output should be combined and only one issue should be created. Also, what happens when the next `benchmark_run` comes in? We don't want to be creating 10 issues for the next 10 benchmark runs that comes in

app/jobs/benchmark_regression_job.rb
tgxworld: https://github.com/ruby-bench/github-api-sandbox/issues/1

I had a look at the above and the details are not meaningful in anyway. Now we know the repo and the commit. But we don't know which `benchmark_type` and for which `benchmark_result_type`. Plus ,  we should probably include how much it has diverted from the standard deviation

app/jobs/benchmark_regression_job.rb
shahsaurabh0605: For the next `benchmark_run` that comes in, it will be checked independently of the previous run whether it is above our given criteria. If it is, then issue has to be opened up.

tgxworld: That is my point. Once we detect a regression, the next n benchmark_runs will trigger the issue until it is fix. That means we will end up with 100 open issues if the regression isn't fixed until 100 runs later.

tgxworld: That is not ideal and is not what we want.

shahsaurabh0605: Do we want to stop the execution of running job once we detect a regression? If so, then when do we resume it. How do we know the regression is fixed?

tgxworld: > Do we want to stop the execution of running job once we detect a regression?
No we don't. We want to think of a way to selectively create an issue. If a similar issue has already been created, we don't want to create a duplicate.

shahsaurabh0605: As you asked, now we are creating an issue per benchmark_run (per `commit`). To selectively creating an issue, what should be the parameter? As in, what does similar issue mean over here?

tgxworld: > As you asked, now we are creating an issue per benchmark_run (per commit) 

Hmm did I mention that somewhere? :thought_balloon: 

> similar issue

When a regression occurs, I'm pretty sure our code will mark it as a regression for the next 10 to 20 benchmark runs of the same result type that comes in. Therefore, we will end up with 10 - 20 issues telling us that a regression has been detected in the benchmark result type

shahsaurabh0605: Ok so, whenever a regression occurs due to a commit, we will have around 10-20 benchmark runs of different categories which will undergo regression. This way we end up with 10-20 issues. So we want only a single regression issue for this. Correct?

tgxworld: Incorrect. Just think in terms of a single benchmark. 

For example:
Commit 1 caused a regression in "activerecord/all" from 5 secs to 10 secs. Issue created.
Commit 2 will still run "activerecord/all" at 10 secs and according to your code. Same issue created.
Commit 2 will still run "activerecord/all" at 10 secs and according to your code. Same issue created.
so on and so forth

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      if benchmark_run.result[category].to_f > results_param && benchmark_run.result[category].to_f <= results_upper_cutoff
+        create_issue benchmark_run, category, results_param
+        @@highest[category] = benchmark_run.result[category].to_f
+        @@flag[category] = true
+      end
+      if @@flag[category] == true && benchmark_run.result[category].to_f < results_param
+        @@highest[category] = 10000
+        @@flag = false
+      end
```
shahsaurabh0605: Ask me if you need any detailed explanation. Not sure if the task can be done this way, but it will satisfy our conditions.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    categories.each do |category|
+      previous_benchmark_results = previous_benchmark_runs.map { |run| run.result[category].to_f }
+
+      results_param = average(previous_benchmark_results) + 2*standard_deviation(previous_benchmark_results)
+      results_upper_cutoff = @@highest[category] - standard_deviation(previous_benchmark_results)
+
+      if benchmark_run.result[category].to_f > results_param && benchmark_run.result[category].to_f <= results_upper_cutoff
+        create_issue benchmark_run, category, results_param
+        @@highest[category] = benchmark_run.result[category].to_f
```
tgxworld: use redis with an expiring key for this please

app/jobs/benchmark_regression_job.rb
tgxworld: Can you explain in detail what you are trying to do? I don't quite follow.

shahsaurabh0605: I am just maintaining a highest variable and a boolean variable. We first set the highest variable to a significantly large value. So, for the first time to detect the regression in benchmark regression(say for `commit x`) which is higher than average + two standard deviations. Then as you said we get next n commits which will undergo benchmark regressions because they have values near values of `commit x`. So I don't consider these commits. 

Now, when the time comes that the regression(caused due to `commit x`) is solved or say the average reaches that point, we again start checking for the commits with normal benchmark regression.

tgxworld: `results_upper_cutoff = @@highest[category] - standard_deviation(previous_benchmark_results)`

What is the significance of `highest` minus one standard deviation? 

What we want is to not open duplicated issues. Will checking via Github's API that an issue for this particular benchmark has not been opened yet before we create a new issue?

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,72 @@
...
+        @@flag[category] = true
```
tgxworld: `category` is not unique to a particular `benchmark_run`

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+      if @@flag[category] == true && benchmark_run.result[category].to_f < results_param
```
tgxworld: `@@flag[category] == true` is not required. Just `@@flag[category]`

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,72 @@
...
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+
+    categories = benchmark_run.result.keys
+
+    current_commit = benchmark_run.initiator
+    previous_commits = Commit.where('created_at < ?', current_commit.created_at).where(repo: current_commit.repo).limit(1000)
+
+    previous_benchmark_runs = []
+    previous_commits.each do |commit|
+      previous_benchmark_runs << BenchmarkRun.find_by(initiator: commit, benchmark_type: benchmark_run.benchmark_type,
```
tgxworld: Not sure how I missed this previously but it is going to hit the database `previous_commits` times

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+
+    categories.each do |category|
+
+      previous_benchmark_results = previous_benchmark_runs.map { |run| run.result[category].to_f }
+
+      results_param = average(previous_benchmark_results) + 2*standard_deviation(previous_benchmark_results)
+
+      if benchmark_run.result[category].to_f > results_param && !check_equal(benchmark_run, category, results_param)
+          create_issue benchmark_run, category, results_param
```
shahsaurabh0605: As you mentioned, I have added methods to check that the same pull request will never opens again. Have a look.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,75 @@
...
+    "Benchmark Regression has been detected:
+    repo: #{benchmark_run.initiator.repo.name}
+    benchmark_category: #{category}
+    benchmark_type: #{benchmark_run.benchmark_type.category}
+    benchmark_result_type: #{benchmark_run.benchmark_result_type.name}
+    Diversion: #{benchmark_run.result[category].to_f - results_param}"
+  end
+
+  def check_equal(benchmark_run, category, results_param)
+    uri = URI.parse(Rails.application.secrets.github_api+"?per_page=100")
```
shahsaurabh0605: I need to set a certain page number here, otherwise it defaults to 30. What should be a good number?

tgxworld: Is it possible to search by date instead? Issues within the last 30days

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,75 @@
...
+    Math.sqrt(previous_benchmark_results.inject(0){|accum, i| accum +(i-results_average)**2 }/(previous_benchmark_results.size - 1).to_f)
+  end
+
+  def request_body(benchmark_run, category, results_param)
+    "Benchmark Regression has been detected:
+    repo: #{benchmark_run.initiator.repo.name}
+    benchmark_category: #{category}
+    benchmark_type: #{benchmark_run.benchmark_type.category}
+    benchmark_result_type: #{benchmark_run.benchmark_result_type.name}
+    Diversion: #{benchmark_run.result[category].to_f - results_param}"
```
tgxworld: Pretty sure the diversion will not be the same and your check will return false.

shahsaurabh0605: Ahh sorry, I need to figure out something about diversion here.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,82 @@
...
+      return true
+    end
+    return false
+  end
+
+  def get_time
+    time_now = Time.now
+    time_before = time_now - (60 * 60 * 24 * 30)
+    time_before_iso = Time.parse((Time.now - (60 * 60 * 24 * 30)).to_s).utc.iso8601
+  end
```
shahsaurabh0605: This method can help us bring all the commits since 30 days.

app/jobs/benchmark_regression_job.rb
shahsaurabh0605: What's the problem if this hits the database `previous_commits` times?

tgxworld: https://secure.phabricator.com/book/phabcontrib/article/n_plus_one/

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    current_commit = benchmark_run.initiator
+    previous_commits = Commit.where('created_at < ?', current_commit.created_at).where(repo: current_commit.repo).limit(1000)
+    previous_benchmark_runs_all = BenchmarkRun.where(initiator: @previous_commits, benchmark_type: benchmark_run.benchmark_type,
+                     benchmark_result_type: benchmark_run.benchmark_result_type)
+
+    previous_benchmark_runs = []
+    previous_benchmark_runs_all.each do |benchmark|
+      previous_benchmark_runs << benchmark
+    end
```
shahsaurabh0605: Done :+1:

tgxworld: You probably do not need another array

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+
+    categories = benchmark_run.result.keys
+
+    current_commit = benchmark_run.initiator
+    previous_commits = Commit.where('created_at < ?', current_commit.created_at).where(repo: current_commit.repo).limit(1000)
+    previous_benchmark_runs = BenchmarkRun.where(initiator: @previous_commits, benchmark_type: benchmark_run.benchmark_type,
+                     benchmark_result_type: benchmark_run.benchmark_result_type)
+
```
shahsaurabh0605: Done :+1:

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,76 @@
...
+  def perform(benchmark_run_id)
+
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+
+    categories = benchmark_run.result.keys
+
+    current_commit = benchmark_run.initiator
+    previous_commits = Commit.where('created_at < ?', current_commit.created_at).where(repo: current_commit.repo).limit(1000)
+    previous_benchmark_runs = BenchmarkRun.where(initiator: @previous_commits, benchmark_type: benchmark_run.benchmark_type,
```
tgxworld: Are tests passing for you locally? `@previous_commits` is probably `nil`.

shahsaurabh0605: Sorry! that was `previous_commits`.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,76 @@
...
+    "Benchmark Regression has been detected:
+    repo: #{benchmark_run.initiator.repo.name}
+    benchmark_category: #{category}
+    benchmark_type: #{benchmark_run.benchmark_type.category}
+    benchmark_result_type: #{benchmark_run.benchmark_result_type.name}
+    Diversion: #{benchmark_run.result[category].to_f - results_param}"
+  end
+
+  def check_equal(benchmark_run, category, results_param)
+    uri = URI.parse(Rails.application.secrets.github_api+"?state=open&since=get_time")
```
tgxworld: `since=get_time` is incorrect because you're passing the string `"get_time"` instead of interpolation.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,76 @@
...
+    response = JSON.parse(Net::HTTP.get(uri))
+    response.each do |response|
+      request_body(benchmark_run, category, results_param).split(/Diversion/).first.eql? response["body"].split(/Diversion/).first
+      return true
```
tgxworld: This method will always return true from the looks of it

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+    return false
+  end
+
+  def get_time
+    time_before_iso = Time.parse((Time.now - (60 * 60 * 24 * 30)).to_s).utc.iso8601
```
tgxworld: This does not take in account of the Time zone configured in the Rails app

config/application.rb
```diff
@@ -14,7 +14,7 @@ class Application < Rails::Application
 
     # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
     # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
-    # config.time_zone = 'Central Time (US & Canada)'
+    config.time_zone = 'Central Time (US & Canada)'
```
shahsaurabh0605: What should I set over here to configure timezone settings?

tgxworld: You don't have to set this. My other comment refers to the fact that Time.now doesn't take into accountof the timezone configures in rails

shahsaurabh0605: So should I use `Time.current`?

tgxworld: Yup

shahsaurabh0605:
Done :+1:

tgxworld:
What is done :question:

shahsaurabh0605:
I have made the changes. You can have a look.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,77 @@
...
+    benchmark_type: #{benchmark_run.benchmark_type.category}
+    benchmark_result_type: #{benchmark_run.benchmark_result_type.name}
+    Diversion: #{benchmark_run.result[category].to_f - results_param}"
+  end
+
+  def check_equal(benchmark_run, category, results_param)
+    uri = URI.parse(Rails.application.secrets.github_api+"?state=open&since=#{get_time}")
+    response = JSON.parse(Net::HTTP.get(uri))
+    response.each do |response|
+      puts response["body"].nil?
```
tgxworld: @shahsaurabh0605 Please check your PR... I've mentioned this before.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,77 @@
...
+    repo: #{benchmark_run.initiator.repo.name}
+    benchmark_category: #{category}
+    benchmark_type: #{benchmark_run.benchmark_type.category}
+    benchmark_result_type: #{benchmark_run.benchmark_result_type.name}
+    Diversion: #{benchmark_run.result[category].to_f - results_param}"
+  end
+
+  def check_equal(benchmark_run, category, results_param)
+    uri = URI.parse(Rails.application.secrets.github_api+"?state=open&since=#{get_time}")
+    response = JSON.parse(Net::HTTP.get(uri))
```
tgxworld: What happens when the HTTP call fails?

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  test "create issue for benchmark regression" do
+    VCR.use_cassette('benchmark_regression') do
+      response = BenchmarkRegressionJob.new.create_issue(@benchmark_run, @benchmark_run.result.keys[0],4.0)
+      assert_equal "201" , response.code
+    end
+  end
+
+  test "check for similar issues" do
+    WebMock.allow_net_connect!
```
tgxworld: This is not what we want to do. You don't really want to hit the actual API when testing because it slows down the tests. Plus, what happens when the actual data makes the test fail?

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    VCR.use_cassette('benchmark_regression') do
+      response = BenchmarkRegressionJob.new.create_issue(@benchmark_run, @benchmark_run.result.keys[0],4.0)
+      assert_equal "201" , response.code
+    end
+  end
+
+  test "check for similar issues" do
+    stub_request(:get, Rails.application.secrets.github_api+"?state=open&since=#{BenchmarkRegressionJob.new.get_time}").
+    with(:headers => {'Accept'=>'*/*', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'User-Agent'=>'Ruby'}).
+    to_return(:status => 200, :body => %Q[["body"]], :headers => {})
```
tgxworld: Is there a reason why we are not using a VCR cassette here :question:

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    assert_equal false, BenchmarkRegressionJob.new.check_equal(@benchmark_run, @benchmark_run.result.keys[0],4.0)
```
tgxworld: There is `assert` and `assert_not` as well.

app/jobs/benchmark_regression_job.rb
tgxworld: I'll prefer we use a percentage rather than a value here.

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,81 @@
...
+    http.use_ssl = true
+    res = http.request(request)
+    raise StandardError, "Request failed with response code #{res.code}" if res.code != "200"
+    response = JSON.parse(res.body)
+    response.each do |response|
+      if request_body(benchmark_run, category, results_param).split(/Diversion/).first.eql? response["body"].split(/Diversion/).first
+        return true
+      end
+    end
+    return false
```
tgxworld: You don't need a return here

tgxworld:
The PR is looking good. One last thing. I need you to fix up your code style because it is all over the place right now. 

1. Leave a blank line before and after for code that spans multi lines.
2. Leave a space after `,` for method parameters. For example: `create_issue(@benchmark_run, @benchmark_run.result.keys[0],4.0)`

test/jobs/benchmark_regression_job_test.rb
shahsaurabh0605: If we are using stub_request then we are not using an actual api request. So do we need vcr cassettes over here?

tgxworld: You are using VCR for the other request to GitHub so I'm curious why we end up stubbing here

shahsaurabh0605: Here we are using current time in the api request which continuously changes. So i think stubbing the request must be the solution.

tgxworld: There is https://github.com/travisjeffery/timecop for that. I just realized you're only unit testing each method individually. How are you sure that when we glue everything together, it'll work?

shahsaurabh0605: To make sure all the pieces glue together, i think i need to add test for the perform method which makes use of all the other methods.

tgxworld: Please do :) I actually think a single test will cover everything. We don't really need to unit test those methods individually. Be sure to cover failure cases as well. Example: `When an issue is not supposed to be created`

test/jobs/benchmark_regression_job_test.rb
```diff
@@ -0,0 +1,26 @@
...
+      )
+    end
+  end
+
+  test "benchmark regression detected" do
+    @benchmark_run.update(result: {"sometime"=>"5.1"})
+    travel_to "2016-06-03 13:50:41 +0530" do
+      VCR.use_cassette('benchmark_regression') do
+        assert_equal ["sometime"], BenchmarkRegressionJob.new.perform(@benchmark_run.id)
+      end
```
shahsaurabh0605: I am a bit stuck here. I am able to travel to a particular time and use VCR cassettes. I need to add two tests, one for the regression detected and issue created and another for issue not created. But how can I test this as the method does not return response code now?

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,75 @@
...
+class BenchmarkRegressionJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(benchmark_run_id)
+    benchmark_run = BenchmarkRun.find_by_id(benchmark_run_id)
+    categories = benchmark_run.result.keys
+    current_commit = benchmark_run.initiator
+    previous_commits = Commit.where('created_at < ?', current_commit.created_at).where(repo: current_commit.repo).limit(1000)
+    previous_benchmark_runs = BenchmarkRun.where(initiator: previous_commits, benchmark_type: benchmark_run.benchmark_type,
+                     benchmark_result_type: benchmark_run.benchmark_result_type)
```
tgxworld: Indentation matters too :wink:

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    categories.each do |category|
+      previous_benchmark_results = previous_benchmark_runs.map { |run| run.result[category].to_f }
+      results_param = average(previous_benchmark_results) + 2*standard_deviation(previous_benchmark_results)
+      if benchmark_run.result[category].to_f > results_param && check_equal(benchmark_run, category, results_param)
+          create_issue benchmark_run, category, results_param
```
tgxworld: This line is not indented consistently

app/jobs/benchmark_regression_job.rb
```diff
@@ -0,0 +1,75 @@
...
+    raise StandardError, "Request failed with response code #{response.code}" if response.code != "201"
+    response
+  end
+
+  def average(previous_benchmark_results)
+    previous_benchmark_results.inject{|sum, el| sum + el}.to_f / previous_benchmark_results.size
+  end
+
+  def standard_deviation(previous_benchmark_results)
+    results_average = average(previous_benchmark_results)
```
tgxworld: You might be able to just calculate the average once of you store it in a variable and pass it to a.methid

test/jobs/benchmark_regression_job_test.rb
tgxworld: Well then you got to think about what you are testing for. In this case, you care that certain HTTP calls will be made under the right scenario. Whether the response code is returned by the method doesn't really matter because the VCR cassette will end up returning the same response code every when is it using the cassette.

tgxworld: Actually just restore the previous unit tests. I realized the trouble of writing integration test isn't worth it.

shahsaurabh0605:
Any errors in indentation?

shahsaurabh0605:
Anything left in this?

tgxworld:
Looks good to me. I'll probably have to merge this over the weekend so that I can fix the tests after merging

tgxworld:
@shahsaurabh0605 Thanks for the work :)

shahsaurabh0605:
Have you seen my reply http://community.rubybench.org/t/gsoc-project-improving-rubybench/99/70 ?

tgxworld:
Closing as stale.

PR #176, opened 2016-08-11
Admin Panel for running benchmarks

shahsaurabh0605:
Added a sample for admin panel for running benchmarks. Currently I have implemented only for the version of releases and not for commits. I am facing the following problem :

Whenever I select the benchmark type and release type the url gets created but it doesn’t showcase any output. How on refreshing the page we will get the required output. How do I solve this?

I have made temporary arrangements to print the output of docker method for now.


tgxworld:
> Whenever I select the benchmark type and release type the url gets created but it doesn’t showcase any output. How on refreshing the page we will get the required output. How do I solve this?

My suggestion is to do without javascript for now since you're not familiar with it. Start with just a form first.

prathamesh-sonpatki:
I agree with Alan. Please work with simple form and then we will think about integrating Javascript.

shahsaurabh0605:
I have made a simple form for `releases` which uses already built-in classes and css. Have a look at it.

app/controllers/admin_controller.rb
```diff
@@ -4,8 +4,43 @@ class AdminController < ApplicationController
...
+   def next
+     if params[:result_type] && params[:admin][:releases_type]
+       @output = `sudo docker run --rm \
+         -e "RUBY_BENCHMARKS=true" \
+         -e "RUBY_MEMORY_BENCHMARKS=true" \
+         -e "RUBY_VERSION= #{params[:admin][:releases_type]}" \
+         -e "API_NAME=#{Rails.application.secrets.api_name}" \
+         -e "API_PASSWORD=#{Rails.application.secrets.api_password}" \
+         -e "INCLUDE_PATTERNS= #{params[:result_type]}"  rubybench/ruby_releases`
+     end
```
shahsaurabh0605: What should I output when I submit the form and go to the next page?

Issue #78, opened 2015-05-17
Add left and right arrow keys to navigate benchmark types


tgxworld:

treble37:
Hi @tgxworld, did you mean to navigate via the keyboard or just by clicking with the mouse in the browser (this is what my PR does)?

tgxworld:
Hi actually I meant the former :)

tgxworld:
I should have written keyboard shortcuts. Apologies for the confusion

treble37:
No worries, how do you feel about inline javascript for this change?  (I'm personally not the biggest fan, but since this is page specific and pretty small....)

tgxworld:
Sure I'm fine with inline this :)

treble37:
Hrrm...the tests were passing locally with `bundle exec rake test` with both Ruby 2.2.1 and 2.3.1....don't suppose you've ever run into something like this before and have any ideas? :)

treble37:
Ok, tests are green @tgxworld and it's working locally toggling between ao_bench and ab_bench

PR #178, opened 2016-08-23
Add left and right arrow keys

treble37:
- Navigate benchmark types
- Addresses #78

### DIFF
--- a/app/views/repos/_form.html.haml
+++ b/app/views/repos/_form.html.haml
@@ -18,3 +18,41 @@
             { selected: @benchmark_run_display_count }, { class: 'form-control' }
 
         .input-group-addon= t('.repos.show.results')
+
+:javascript
+  $(document).on('ready page:load', function() {
+    //Add left and right arrow keys to navigate benchmark types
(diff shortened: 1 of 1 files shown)


app/views/repos/_form.html.haml
```diff
@@ -1,3 +1,39 @@
+:javascript
+  //Add left and right arrow keys to navigate benchmark types
+  var opts = $('.result-types-form select.form-control option');
```
tgxworld: We have to make sure that the document is loaded before executing the JavaScript code. You can have a look at how it is done in https://github.com/ruby-bench/ruby-bench-web/blob/3ac72884f045d112803e0370fdb79dfd9d519c16/app/views/repos/_show_chart.html.haml#L27-L30 :)

Also we should move the JavaScript code to the bottom of the file so that it gets loaded last.

treble37:
ok @tgxworld - tests are green...what do you think?

tgxworld:
Thanks @treble37  :) Let's try this out

PR #180, opened 2016-08-27
Make seeds.rb closer to idempotent

kbrock:
I tend to start up projects using:

```
bundle exec rake db:create db:migrate db:seed db:schema:dump db:test:prepare
```

This fails since the `BenchmarkRun` model has been setup before the field `benchmark_type_id` was added to the schema:

```
ActiveRecord::UnknownAttributeError: unknown attribute 'benchmark_type_id' for BenchmarkRun.
```

before
-------

If I try to run `rake db:seed` again, it will blow complain about the records already existing.

after
-----

It lets me run `rake db:seed` again.

While this change does not make the `db:seeds` completely idempotent, it does allow it be be run again successfully and duplicates minimal data.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -6,41 +6,43 @@
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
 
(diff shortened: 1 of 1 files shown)


tgxworld:
Thanks!

PR #182, opened 2016-12-06
Fix 'view-benchmark-graph-test.rb' timing out

alahab2:
This change fixes the issue of `view-benchmark-graph-test.rb` timing out:
- Add 'capybara-screenshot'
- Fix database_cleaner to reset the DB properly
- Fix timimg out of "User should be able to select number of benchmark runs to display for long running benchmark graphs"

### DIFF
--- a/test/acceptance/test_helper.rb
+++ b/test/acceptance/test_helper.rb
@@ -8,7 +8,7 @@ class AcceptanceTest < ActionDispatch::IntegrationTest
 
   self.use_transactional_tests = false
 
-  DatabaseCleaner.strategy = :truncation
+  DatabaseCleaner.strategy = :deletion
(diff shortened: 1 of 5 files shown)


test/acceptance/test_helper.rb
```diff
@@ -8,7 +8,7 @@ class AcceptanceTest < ActionDispatch::IntegrationTest
 
   self.use_transactional_tests = false
 
-  DatabaseCleaner.strategy = :truncation
+  DatabaseCleaner.strategy = :deletion
```
tgxworld: Why does this need to be changed? :thinking:

alahab2: Without this the database doesn't get cleaned properly. You can try locally with :truncation, it doesn't work.

PR #183, opened 2017-01-27
Possible typo in database seed benchmark name

jerryliu55:
In development, running `bundle exec rake db:setup` with the current `db/seeds.rb` causes the following (a result from `ao_bench` is displayed in the graph for `ab_bench`):

image

The reverse (a result from `ab_bench` is displayed in the graph for `ao_bench`) is true as well.  
This PR resolves this problem, as follows:

image

If this is intended behaviour, then feel free to ignore this PR.


tgxworld:
Good catch :+1:

PR #185, opened 2017-02-01
Add option to run liquid benchmarks

jerryliu55:
Run liquid benchmarks with ruby benchmarks using an option


tgxworld:
Looks good :+1:  Thank you!

PR #187, opened 2017-02-07
Fix footer links so that they're centered

jerryliu55:
The footer links on each page were off-centre. This centres them again.

before:
image

after:
image


tgxworld:
Thank you!!

PR #192, opened 2017-02-24
Add JSON format by changing ChartBuilder

jerryliu55:
I believe this is a better and cleaner way to expose a JSON format

## Changes to `ChartBuilder`
Previously, the `ChartBuilder` object (app/services/chartbuilder.rb) was made mainly to help with visual charts (ie. by storing HTML version strings in `@data[:categories]` - not generating charts in JSON format.

This changes `ChartBuilder` to
* store a versions array in `@data[:categories]`, from which both JSON versions as well as HTML versions can be derived
* store the `benchmark_result_type` (app/models/benchmark_result_type.rb) of that chart, making it clearer that the result type is part of the chart
* add a function `construct_from_cache` which takes in cached `@data[:categories]` and `@data[:columns]` and returns a `ChartBuilder` object from which visual and JSON charts can be generated

## Changes to `RepoController`
The `@charts` object in `RepoController` previously was an array of pairs of `[ChartBuilder#@data, BenchmarkResultType]`

* change `@charts` to be an array of `ChartBuilder` objects, since we internalized the `benchmark_result_type`
* the [block](https://github.com/Shopify/ruby-bench-web/compare/master...Shopify:add-json-format-clearer?expand=1#diff-a75f08af70f7e9f0da0995bdf9413e30R46) we pass to `chart_builder.build_columns` now returns a version hash rather than an HTML string, which gets appended to `chart_builder.data[:categories]`

## JSON format
example: https://gist.github.com/jerryliu55/e89e007386018977b6b8fa87c9e71f0f

* decided to separate the versions from the datapoints since versions are shared across charts. This way, versions won't be repeated if there is more than one chart

@tgxworld @kirs


test/integration/repos_test.rb
```diff
@@ -12,4 +12,87 @@ class ReposTest < ActionDispatch::IntegrationTest
...
+    assert_includes res.keys, :benchmark_name
+    assert_includes res.keys, :charts
+    assert_includes res.keys, :versions
+    assert_equal res[:charts].length, 1
+    assert_includes res[:charts][0].keys, :measurement
+    assert_includes res[:charts][0].keys, :unit
+    assert_equal res[:charts][0][:datasets].length, 1
+  end
+
+  test "JSON generation works when there are two charts" do
```
kirs: Good call on adding controller level tests that reproduce the environment and check the JSON.

test/services/chart_builder_test.rb
```diff
@@ -4,21 +4,22 @@ class ChartBuilderTest < ActiveSupport::TestCase
...
-    chart_builder = ChartBuilder.new([benchmark_run, other_benchmark_run])
+    chart_builder = ChartBuilder.new([benchmark_run, other_benchmark_run], benchmark_run.benchmark_result_type)
 
-    chart_columns = chart_builder.build_columns do |benchmark_run|
-      "Commit: #{benchmark_run.initiator.sha1}"
+    chart_data = chart_builder.build_columns do |benchmark_run|
+      { commit: benchmark_run.initiator.sha1 }
     end
 
     assert_equal(
```
kirs: Compared to massive changes introduced to `ChartBuilder`, I think it makes sense to also extend the coverage in `ChartBuilderTest`. Controller tests are great to have, but it's not enough.

app/services/chart_builder.rb
```diff
@@ -1,5 +1,38 @@
...
+  #   {
+  #     version: "0",
+  #     environment: "ruby 2.2.0dev"
+  #   },
+  #   {
+  #     version: "0",
+  #     environment: "ruby 2.2.0dev"
+  #   }
+  # ]
+  attr_accessor :data
```
kirs: If you know that the only two keys will be `columns` and `categories`, then why not 

```
attr_accessor :columns, :categories
```

You'd usually like to avoid variable and method names like `data`. Everything in the computer world is data and such names are not telling you what do they refer to.

jerryliu55:
@kirs made the changes

One dataset represents the benchmark results across many commits/versions. The `versions` array in the JSON format corresponds to these commits/versions, in the same order. One has to find the index in the versions array and use that to look up the corresponding value in the datasets.

The reason for this format is that a single benchmark (ex. discourse/home) can have multiple datasets representing the different charts and variations in each chart. All of the datasets share the same `versions` array, so I pulled `versions` out of the datasets.

app/controllers/repos_controller.rb
```diff
@@ -92,33 +89,27 @@ def show_releases
...
-              temp << "#{key}: #{value}<br>"
-            end
+          version = { version: benchmark_run.initiator.version }
 
-            environment = temp
+          # If there is more information about the environment, we add it to `version` 
+          if environment.is_a?(Hash)
+            version.merge!(environment)
+          else
+            #   a new key in `version`
```
kirs: ✂️

app/helpers/charts_helper.rb
```diff
@@ -0,0 +1,27 @@
+module ChartsHelper
+  # Generate an HTML string representing `version`, with each pair on a new line
+  #
+  # `version` will usually look like:
+  # {
+  #   commit: "b6589fc",
```
kirs: `commit_sha`?

app/helpers/charts_helper.rb
```diff
@@ -0,0 +1,27 @@
+module ChartsHelper
+  # Generate an HTML string representing `version`, with each pair on a new line
```
kirs: 👍  for docs

app/services/chart_builder.rb
```diff
@@ -1,35 +1,63 @@
 class ChartBuilder
-  def initialize(benchmark_runs)
+  
+  # @columns is JSON that looks like [{ name: "benchmark1", data: [1.1, 1.2] }]
```
kirs: why `@columns` is JSON? The ruby class should operate with Ruby data structures. JSON is just a side effect of using serialization

app/views/repos/_script.html.haml
```diff
@@ -5,5 +5,6 @@
     = link_to benchmark_type.github_url, class: 'btn btn-primary', target: '_link' do
       = t('repos.view_github')
 
+  - require 'net/http'
```
kirs: Why would you need this? It's run in the context of Rails where net/http is already required.

test/integration/repos_test.rb
```diff
@@ -12,4 +12,87 @@ class ReposTest < ActionDispatch::IntegrationTest
       get '/tgxworld/rails'
     end
   end
+
+  test "JSON generation works when there is one chart" do
```
kirs: Should we also test the part of the view when `@charts.empty?` is true?

test/services/chart_builder_test.rb
```diff
@@ -4,21 +4,83 @@ class ChartBuilderTest < ActiveSupport::TestCase
...
+
+  test "#build_columns returns self" do
+    benchmark_run = create(:commit_benchmark_run, result: { 'some_time' => 5, 'some_other_time' => 5 })
+    chart_builder = ChartBuilder.new([benchmark_run], benchmark_run.benchmark_result_type)
+
+    chart = chart_builder.build_columns do |benchmark_run|
+      { commit: benchmark_run.initiator.sha1 }
+    end
+
+    assert chart.instance_of?(ChartBuilder)
```
kirs: `assert_instance_of`.

You may also want to check that the returned object contains legit data

test/services/chart_builder_test.rb
```diff
@@ -4,21 +4,83 @@ class ChartBuilderTest < ActiveSupport::TestCase
...
+    cache_read = { 
+      datasets: [{ name: "benchmark1", data: [1.2] }],
+      versions: [{ version: 1 }]
+    }
+
+    benchmark_result_type = { measurement: "Execution time", unit: "Seconds" }
+
+    chart_builder = ChartBuilder.construct_from_cache(cache_read, benchmark_result_type)
+
+    assert chart_builder.instance_of?(ChartBuilder)
```
kirs: Same thing about assert_instance_of. 
Let's assert the data that the returned `chart_builder` contains

kirs:
@tgxworld what do you think about the proposed JSON format (https://gist.github.com/jerryliu55/e89e007386018977b6b8fa87c9e71f0f)?

test/integration/repos_test.rb
jerryliu55: done :+1:

test/services/chart_builder_test.rb
jerryliu55: ok, i pulled these assertions into the other tests

app/controllers/repos_controller.rb
```diff
@@ -92,33 +88,26 @@ def show_releases
...
 
-    respond_to do |format|
-      format.html do
-        @result_types = fetch_categories
-      end
-
-      format.js
-    end
+    @result_types = fetch_categories if request.format.html?
+    @benchmark_name = params[:result_type]
```
kirs: If the user input (`params[:result_type]`) is an array, hash or 1GB file upload, is it going to be serialized to JSON and rendered? That sounds like a security vulnerability. 

You'd like to check if the value looks legit. In this case, it's enough to do `@benchmark_name = params[:result_type].to_s` to convert it into string and don't care about the actual type.

app/helpers/charts_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+  # ruby 2.2.0dev
+  def chart_version_to_html(version)
+    version.map do |k, v|
+      if k == :environment
+        v
+      else
+        new_key = k.to_s.split('_').map(&:capitalize).join(' ')
+        "#{new_key}: #{v}" 
+      end
+    end.join("<br>")
```
kirs: AFAIK the valid way of br is `<br/>`

app/helpers/charts_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+  # Commit: b6589fc
+  # Commit Date: 2017-02-23T16:20:13.338Z
+  # Commit Message: fix something
+  # ruby 2.2.0dev
+  def chart_version_to_html(version)
+    version.map do |k, v|
+      if k == :environment
+        v
+      else
+        new_key = k.to_s.split('_').map(&:capitalize).join(' ')
```
kirs: There's a `titleize` method in Rails exactly for that:

```ruby
'raiders_of_the_lost_ark'.titleize  # => "Raiders Of The Lost Ark"
```

The next line can be shortened into `"#{k.to_s.titleize}: #{v}"`

app/models/repo.rb
```diff
@@ -25,12 +25,11 @@ def generate_sparkline_data
...
 
-        chart_builder = ChartBuilder.new(benchmark_runs.sort_by do |benchmark_run|
-          benchmark_run.initiator.created_at
-        end)
+        runs = benchmark_runs.sort_by { |run| run.initiator.created_at }
+        chart_builder = ChartBuilder.new(runs, benchmark_result_type).build_columns
 
         charts[benchmark_type.category] ||= []
-        charts[benchmark_type.category] << [benchmark_result_type.name, chart_builder.build_columns]
+        charts[benchmark_type.category] << [benchmark_result_type.name, { columns: chart_builder.columns.to_json, categories: chart_builder.categories }]
```
kirs: It doesn't sound right that the Ruby class `Repo` returns data structure serialized in JSON. If you need it to be JSON on the client side, it's controller business to serialize data into JSON.

app/services/chart_builder.rb
```diff
@@ -1,35 +1,63 @@
...
+  # {
+  #   name: "Memory used",
+  #   unit: "Bytes"
+  # }
+  attr_reader :benchmark_result_type
+
+  # `cache_read` looks like 
+  # { datasets: [{ name: "benchmark1", data: [1.1, 1.2] }], versions: [version_hash, version_hash] }
+  def self.construct_from_cache(cache_read, benchmark_result_type)
+    chart_builder = ChartBuilder.new([], benchmark_result_type)
```
kirs: You don't need to reference the class name from itself. You can do `self.class.new` here

test/integration/repos_test.rb
```diff
@@ -12,4 +12,103 @@ class ReposTest < ActionDispatch::IntegrationTest
...
+  end
+
+  test "#JSON generation works when there are no charts" do
+    benchmark_type = create(:benchmark_type)
+    benchmark_result_type = create(:benchmark_result_type)
+    repo = benchmark_type.repo
+    org = repo.organization
+
+    get "/#{org.name}/#{repo.name}/commits.json?result_type=#{benchmark_type.category}",
+      params: { display_count: 2, result_type: benchmark_result_type }
```
kirs: You pass `result_type` as a query param (`result_type=#{benchmark_type.category}"`) and as a form param (`result_type: benchmark_result_type`). Which one do you actually want to pass here? Now at least one of them will be ignored.

test/services/chart_builder_test.rb
```diff
@@ -4,21 +4,63 @@ class ChartBuilderTest < ActiveSupport::TestCase
...
+    chart_builder = ChartBuilder.new([benchmark_run, other_benchmark_run], benchmark_run.benchmark_result_type)
 
-    chart_columns = chart_builder.build_columns do |benchmark_run|
-      "Commit: #{benchmark_run.initiator.sha1}"
+    chart_builder2 = chart_builder.build_columns do |benchmark_run|
+      { commit: benchmark_run.initiator.sha1 }
     end
 
+    # check that `build_columns` returns self
+    assert_equal chart_builder, chart_builder2
```
kirs: If you want to compare that objects are the same, it has to be `assert_equal chart_builder.object_id, chart_builder2.object_id`. You also won't need the comment because comparing `object_id` tells developers exactly what the comment is trying to tell

app/helpers/charts_helper.rb
jerryliu55: It used to use `<br>` so I just used that: https://github.com/ruby-bench/ruby-bench-web/blob/master/app/controllers/repos_controller.rb#L58

app/models/repo.rb
jerryliu55: what about doing it in the view? Similar to how we call the `chart_version_to_html` method

kirs: Even better!

app/controllers/repos_controller.rb
```diff
@@ -27,53 +27,53 @@ def show
...
+            # this return values gets appended to ChartBuilder#@versions
+            version
           end
 
-          $redis.set(cache_key, columns.to_json)
-          [columns, benchmark_result_type]
+          $redis.set(cache_key, {
+            datasets: JSON.parse(chart_builder.data[:columns]),
+            versions: chart_builder.data[:categories]
+          }.to_json)
```
tgxworld: @kirs Good call on msgpack. I've actually never known about it before :+1:

tgxworld:
@jerryliu55 Thank you for all the work! Lets try this out and I'll be deploying this soon :)

Thanks @kirs for reviewing :+1:

tgxworld:
This has been deployed :) Thank you all!

kirs:
@tgxworld thanks for deploying it!

jerryliu55:
Thanks @tgxworld! Note that the cache will have to be flushed when this is deployed since we are changing how the cache store/read works.

tgxworld:
@jerryliu55 Ok done :+1:

Gemfile.lock
```diff
@@ -327,4 +334,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.13.6
+   1.14.5
```
WaKeMaTTa: @tgxworld this needs to be rolled back if you are using Heroku. ;)

kirs: I'm curious what's wrong with Bundler 1.14.5 on Heroku.

WaKeMaTTa:
Nothing, but heroku recomend 1.13 because he use 1.13
https://devcenter.heroku.com/articles/bundler

jerryliu55:
@tgxworld just wondering if this has been deployed properly? when I go to `https://rubybench.org/ruby/ruby/commits.json?result_type=app_mandelbrot&display_count=2000`, I get a `{"status":406,"error":"Not Acceptable"}`

tgxworld:
Yup the changes have been deployed. Hmm I'm getting the same error too.

tgxworld:
@jerryliu55 Oops it seems like I made a mistake. There changes were not deploy at all because I forgot to merge in master to production branch. Can you try again?

jerryliu55:
Looks like it's working, thanks for figuring it out! :) I guess we should probably paginate the results, since right now it returns all 2000 datapoints

PR #199, opened 2017-05-18
Compare benchmarks on commits graph

bmarkons:
New feature: Compare graphs for two different benchmarks.

Take a closer look at [discussion](https://community.rubybench.org/t/rewriting-the-scope-all-benchmark-in-sequel-and-raw/138/12).


bmarkons:
Here's what it looks like: 

image

This way you choose benchmark and select any other which you want to compare with. Comparing graph will be displayed only if selected benchmark is the same type (e.g. Execution time).

UI has some issues and it's far from perfect but we have something to work with and something valuable is being displayed.

@tgxworld pls take a look. If you agree, we can merge it and make it better in next PR :bowing_man: 

Right now, I want to get sequel benchmarks displayed so we can compare them with activerecord.

app/models/benchmark_run.rb
```diff
@@ -14,7 +14,7 @@ class BenchmarkRun < ApplicationRecord
   # FIXME: Remove this and order by Commit#created_at
   default_scope { order("#{self.table_name}.created_at DESC")}
 
-  scope :fetch_commit_benchmark_runs, ->(form_result_type, benchmark_result_type, limit) {
+  scope :fetch_commit_benchmark_runs, -> (form_result_type, benchmark_result_type, limit) {
```
noahgibbs: Really minor thing: if you're just making a one-space change in a given file, maybe best to skip it for now :-)

bmarkons: Oh, yeah. It's irrelevant to this PR and only making noise :+1:

app/services/chart_builder.rb
```diff
@@ -31,33 +31,71 @@ def self.construct_from_cache(cache_read, benchmark_result_type)
...
+      end
+
+      @columns = new_columns
+      self
     end
+  end
 
-    @columns = new_columns
-    self
+  def chart_version_to_html(version)
```
noahgibbs: This kind of thing is usually better to put in a helper. Anything with HTML output, basically.

app/views/repos/_script.html.haml
```diff
@@ -1,9 +1,10 @@
...
+- if script_url = benchmark.script_url
   %h2.l-align-center
     = t('repos.script_title', graph_name: benchmark_name)
 
-    = link_to benchmark_type.github_url, class: 'btn btn-primary', target: '_link' do
+    = link_to benchmark.github_url, class: 'btn btn-primary', target: '_link' do
       = t('repos.view_github')
 
   - formatter = Rouge::Formatters::HTMLPygments.new(Rouge::Formatters::HTML.new)
+  - require 'net/http' unless defined?(Net::HTTP)
```
noahgibbs: Require will make sure that net/http isn't already required, normally. Also, you normally want to do your require elsewhere, not directly in the view.

bmarkons: I had an issue with net/http locally saying that Net::HTTP is not recognized, although not having the same in production and test environment. So this was a quick workaround. 

Do you have any idea if I can maybe avoid this require? Since it's needed only in dev environment.

Where should I place `require`'s, is there any good practice?

noahgibbs: It would be much better to put it in the controller and/or tests that use this view. You *could* also put it somewhere under config, such as config/initializers, which would guarantee it was *always* loaded. Might be too much, might not. If you only need a thing in the dev environment, you can literally put an "if" around it, such as "if Rails.env.development?"

bmarkons: Awesome, thanks @noahgibbs.

app/controllers/repos_controller.rb
```diff
@@ -105,26 +60,144 @@ def releases
...
+
+      chart
+    else
+      nil
+    end
+  end
+
+  def benchmark_runs_for(benchmark_type)
+    BenchmarkRun
+    .fetch_commit_benchmark_runs(@benchmark.category, benchmark_type, @display_count)
```
tgxworld: Style nitpick: I prefer to have this indented

test/acceptance/compare_benchmarks_test.rb
```diff
@@ -0,0 +1,71 @@
...
+      benchmark_type: @sequel_scope_all,
+      initiator: @sequel_commit,
+      benchmark_result_type: @memory_benchmark
+    )
+
+    @commit_dates = [@rails_commit.created_at, @sequel_commit.created_at].map{ |date| date.strftime("%Y-%m-%d") }
+  end
+
+  test "User should be able to compare benchmarks across repos" do
+    visit commits_path(@rails_org.name, @rails_repo.name, result_type: @active_record_scope_all.category, compare_with: @sequel_scope_all.category)
```
tgxworld: The line is alittle too long here. Let's indent the hash param :)

app/controllers/repos_controller.rb
```diff
@@ -1,6 +1,17 @@
 class ReposController < ApplicationController
-  before_action :find_organization_by_name
-  before_action :find_organization_repo_by_name
+
+  if Rails.env.development?
+    require "net/http"
+  end
```
WaKeMaTTa:
```ruby
require 'net/http" if Rails.env.development?
```

app/controllers/repos_controller.rb
WaKeMaTTa: Yes! please

```ruby
BenchmarkRun
  .fetch_commit_benchmark_runs(@benchmark.category, benchmark_type, @display_count)
  .sort_by { |run| run.initiator.created_at }
```

app/controllers/repos_controller.rb
```diff
@@ -105,26 +60,144 @@ def releases
...
+  def benchmark_runs_for(benchmark_type)
+    BenchmarkRun
+    .fetch_commit_benchmark_runs(@benchmark.category, benchmark_type, @display_count)
+    .sort_by { |run| run.initiator.created_at }
+  end
+
+  def comparing_runs_for(benchmark_type)
+    BenchmarkRun
+    .fetch_commit_benchmark_runs(@comparing_benchmark.category, benchmark_type, @display_count)
+    .sort_by { |run| run.initiator.created_at }
```
WaKeMaTTa: same

app/models/benchmark_type.rb
```diff
@@ -1,5 +1,6 @@
 class BenchmarkType < ApplicationRecord
   default_scope { order("#{self.table_name}.category ASC") }
+  scope :all_except, -> (user) { where.not(id: user.id) }
```
WaKeMaTTa: Not need it to pass `id` rails is inteligent.

```ruby
scope :all_except, -> (user) { where.not(id: user) }
```

app/controllers/repos_controller.rb
bmarkons: :+1:  thanks @WaKeMaTTa

bmarkons:
I am struggling with one test and can't find out why it fails from time to time :sweat:

https://travis-ci.org/ruby-bench/ruby-bench-web/builds/238419839#L593

app/assets/javascripts/initialize.js
```diff
@@ -1,11 +1,66 @@
...
+
+  var organizationName = $resultTypesForm.data('organization-name');
+  var repoName = $resultTypesForm.data('repo-name');
+  var name = $resultTypesForm.data('name');
+
+  var resultType = $('#benchmark_run_benchmark_type').val() || "";
+  var benchmarkRunDisplayCount = $('#benchmark_run_display_count').val();
+  var compareWithBenchmark = $('#benchmark_run_compare_with').val();
+
+  if(compareWithBenchmark){
```
robin850: Pretty nit-picky but maybe you could add surrounding spaces around braces.

app/assets/javascripts/initialize.js
```diff
@@ -28,11 +83,11 @@ $(document).on('turbolinks:load', function() {
     switch(e.which) {
       case 37:  //left
         optionIndex = incrementOptionIndex(optionIndex, optsLength, -1);
-        break;
```
robin850: Also nit-picky but it seems more idiomatic to align the `break` with the block's content in JavaScript.

app/services/chart_builder.rb
```diff
@@ -31,33 +31,61 @@ def self.construct_from_cache(cache_read, benchmark_result_type)
...
 
   def build_columns
-    @benchmark_runs.each do |benchmark_run|
-      if block_given?
-        version = yield(benchmark_run)
-        @categories ||= []
-        @categories << version if version != @categories.last
+    if @comparing_runs.present?
+      (@benchmark_runs + @comparing_runs)
+      .sort_by{ |benchmark_run| benchmark_run.initiator.created_at }
```
robin850: As @tgxworld suggested earlier, you could indent this by a level.

Also, to be honest I didn't test it but there should be a way to make this method a little bit easier to read. What do you think about something like:

~~~ruby
def build_columns
  columns = {}

  if @comparing_runs.present?
    runs = (@benchmark_runs + @comparing_runs)
             .sort_by { |run| run.initiator.created_at }
  else
    runs = @benchmark_runs
  end
    
  runs.each do |run|
    version = nil
    if block_given?
      version = yield(run)
      @categories ||= []
      @categories << version if version != @categories.last
    end

    run.result.each do |key, value|
      if @comparing_runs.present?
        columns["#{key}_#{run.benchmark_type.category}"] ||= []
        columns["#{key}_#{run.benchmark_type.category}"] << [version, value.to_f]
      else
        columns[key] ||= []
        columns[key] << value.to_f
      end
    end
  end

  @columns = columns.map do |name, data|
    { name: name, data: data}
  end

  self
end
~~~

SamSaffron:
@bmarkons  @robin850  had some comments, can you be sure to address them and the merge conflict, I really want this merged into the repo and on the website as soon as possible. 

We can refine it further in future PRs

bmarkons:
sure @SamSaffron, I'm on it.

bmarkons:
Thanks @robin850 :) I've made changes you've suggested.

robin850:
@bmarkons : No problem, good job. Does the timeout change is related to the refactoring of the `build_columns` method ? The code I've given may be terribly inefficient.

bmarkons:
No, that change doesn't affect performance. The code I've written in this PR is pretty much inefficient but I plan to make it better in next iteration.

robin850:
Okay great, this can be shipped then ! 🙂 🎉

PR #203, opened 2017-06-02
Handle only commits to main branches

bmarkons:
@tgxworld I guess we only want to track push events on master/trunk branches, so we need to check if push occured to main branch.

Note: we didn't need this check on fork since we had events only on main branch.


bmarkons:
need to add few tests before ready to merge.

bmarkons:
It's ready now :shipit:

tgxworld:
Awesome :+1:

PR #204, opened 2017-06-02
Execute scripts on bare metal server

bmarkons:


app/jobs/remote_server_job.rb
```diff
@@ -20,53 +33,47 @@ def perform(initiator_key, benchmark, options = {})
...
-          -e \"OPTCARROT_BENCHMARK=#{options[:optcarrot_benchmarks]}\"
-          -e \"LIQUID_BENCHMARK=#{options[:liquid_benchmarks]}\"
-          -e \"RUBY_COMMIT_HASH=#{commit_hash}\"
-          -e \"API_NAME=#{Rails.application.secrets.api_name}\"
-          -e \"API_PASSWORD=#{Rails.application.secrets.api_password}\"
-          #{build_include_patterns(options[:include_patterns])}
-          rubybench/ruby_trunk".squish
-      ]
-    )
+    ruby = true
```
noahgibbs: I assume these params will be set in some different way when you're farther along...

bmarkons: yeah I guess. You have script which takes arguments but on the other side you are calling it always with the same, currently.

I moved those values to variables so it could be somehow easier to tell which value stands for which argument.

app/jobs/remote_server_job.rb
```diff
@@ -13,59 +26,39 @@ def perform(initiator_key, benchmark, options = {})
       password: secrets.bare_metal_server_password
     ) do |ssh|
 
-      send(benchmark, ssh, initiator_key, options)
+      send(benchmark_group, ssh, initiator_key, options)
```
bmarkons: @noahgibbs What do you think about calling method like this? Is it a good approach?

noahgibbs: I think it's perfectly reasonable. But I'm also very pro-metaprogramming :-)

bmarkons:
Since https://github.com/ruby-bench/ruby-bench-docker/pull/26 is merged, according to me this one is mergable. :ship:

app/jobs/remote_server_job.rb
```diff
@@ -231,4 +138,24 @@ def ssh_exec!(ssh, command)
...
+    Rails.application.secrets.bare_metal_server_password
+  end
+
+  def api_name
+    Rails.application.secrets.api_name
+  end
+
+  def api_password
+    Rails.application.secrets.api_password
+  end
```
bmarkons: @noahgibbs Do you think these 5 methods above are too much?

Maybe better way would be to make new module `Secrets` and move them there. That way whenever I need secrets I would include `Secrets` module.

bmarkons:
Since https://github.com/ruby-bench/ruby-bench-docker/pull/26 is merged, according to me this one is mergable. :ship:

bmarkons:
Since https://github.com/ruby-bench/ruby-bench-docker/pull/26 is merged, according to me this one is mergable. :ship:

app/jobs/remote_server_job.rb
tgxworld: Yea I think aren't really necessary. Perhaps extract `Rails.application.secrets` into a private method is enough?

test/jobs/remote_server_job_test.rb
```diff
@@ -109,160 +77,91 @@ class RemoteServerJobTest < ActiveJob::TestCase
...
+    bundler
+    sequel
+  end
+
+  def ruby
+    @ruby = true
+    @memory = true
+    @optcarrot = true
+    @liquid = true
+    @ruby_trunk_script = RemoteServerJob::RUBY_TRUNK
```
tgxworld: This is a little odd for me. Why not just reference the constant directly.

bmarkons: yeah, you're right. It doesn't make sense to me neither :confused: 
thanks!

tgxworld:
Are we cloning ruby-bench-docker into the bare metal server?

tgxworld:
Merged in 9bb7b392d7bed6c7866293e25d73165b2cb42ace

tgxworld:
Thank you @bmarkons

bmarkons:
cool :+1: :tada:

PR #205, opened 2017-06-05
Introduce rubocop

bmarkons:
Why not automate style check? :policeman:

Rupocop is currently configured with `.rubocop.yml` checked out from https://github.com/rails/rails.

Though I've modified `Style/StringLiterals` to single quotes and commented out `Layout/IndentationConsistency`.

I ran `rubocop --auto-correct` so thats why all those files have been changed.

WDYT?


.rubocop.yml
```diff
@@ -0,0 +1,125 @@
...
+Layout/SpaceInsideHashLiteralBraces:
+  Enabled: true
+
+Layout/SpaceInsideParens:
+  Enabled: true
+
+# Check quotes usage according to lint rule below.
+Style/StringLiterals:
+  Enabled: true
+  EnforcedStyle: single_quotes
```
bmarkons: The most of project uses `'` instead of `"` for string literals. So I've enforced `'` though I would rather go with `"`.

What do you suggest, should I leave single quotes or switch to double quotes? :slightly_smiling_face:

Anyway we should be consistent with using quotes.

tgxworld: I actually rather not have to enforce it. To me it doesn't make a difference even if `"` is used without any interpolation.

bmarkons: I think it's good to be consistent with style, though it's not that crucial thing :)

noahgibbs:
Thanks for doing the --autocorrect in a separate commit :-) Looks fine to me.

Gemfile
```diff
@@ -20,6 +20,7 @@ gem 'coffee-rails'
 gem 'responders'
 gem 'jbuilder'
 gem 'msgpack'
+gem 'rubocop', require: false
```
tgxworld: Do we need rubocop in production?

PR #210, opened 2017-06-08
Add manual runner

bmarkons:
I needed to run Sequel benchmarks for last 2000 commits. So far I could do it in production console in a pretty hacky (unsafe) way. Therefore I want to add `app/services/manual_runner.rb` which would fetch, store and run last 2000 commits from github api. That way we could run last number of commits manually in a tested way.

### DIFF
--- a/app/services/commits_runner.rb
+++ b/app/services/commits_runner.rb
@@ -1,7 +1,6 @@
-module CommitsRunner
-  module_function
+class CommitsRunner
 
-  def run(commits)
+  def self.run(commits)
     commits.each do |commit|
       if(valid?(commit))
         create(commit)
@@ -10,11 +9,13 @@ module CommitsRunner
     end
   end
 
-  def valid?(commit)
+  private
+
+  def self.valid?(commit)
     !Commit.merge_or_skip_ci?(commit[:message]) && Commit.valid_author?(commit[:author_name])
   end
 
-  def create(commit)
+  def self.create(commit)
     Commit.find_or_create_by(sha1: commit[:sha]) do |c|
       c.url = commit[:url]
       c.message = commit[:message]
@@ -22,4 +23,5 @@ module CommitsRunner
       c.created_at = commit[:created_at]
     end
   end
+
 end
--- a/app/services/commits_runner.rb
+++ b/app/services/commits_runner.rb
@@ -1,0 +1,25 @@
+module CommitsRunner
+  module_function
+
+  def run(commits)
+    commits.each do |commit|
+      if(valid?(commit))
+        create(commit)
+        BenchmarkPool.enqueue(commit[:repo][:name], commit[:sha])
+      end
+    end
+  end
+
+  def valid?(commit)
+    !Commit.merge_or_skip_ci?(commit[:message]) && Commit.valid_author?(commit[:author_name])
+  end
+
+  def create(commit)
+    Commit.find_or_create_by(sha1: commit[:sha]) do |c|
+      c.url = commit[:url]
+      c.message = commit[:message]
+      c.repo_id = commit[:repo][:id]
+      c.created_at = commit[:created_at]
+    end
+  end
+end
--- a/app/services/commits_runner.rb
+++ b/app/services/commits_runner.rb
@@ -1,9 +1,9 @@
 class CommitsRunner
   def self.run(commits)
     commits.each do |commit|
-      if (valid?(commit))
+      if valid?(commit)
         create(commit)
-        BenchmarkPool.enqueue(commit[:repo][:name], commit[:sha])
+        BenchmarkPool.enqueue(commit[:repo].name, commit[:sha])
       end
     end
   end
@@ -15,10 +15,10 @@ class CommitsRunner
   end
 
   def self.create(commit)
-    Commit.find_or_create_by(sha1: commit[:sha]) do |c|
+    Commit.find_or_create_by!(sha1: commit[:sha]) do |c|
       c.url = commit[:url]
       c.message = commit[:message]
-      c.repo_id = commit[:repo][:id]
+      c.repo_id = commit[:repo].id
       c.created_at = commit[:created_at]
     end
   end
--- a/app/services/commits_runner.rb
+++ b/app/services/commits_runner.rb
@@ -1,8 +1,7 @@
 class CommitsRunner
-
   def self.run(commits)
     commits.each do |commit|
-      if(valid?(commit))
+      if (valid?(commit))
         create(commit)
         BenchmarkPool.enqueue(commit[:repo][:name], commit[:sha])
       end
@@ -23,5 +22,4 @@ class CommitsRunner
       c.created_at = commit[:created_at]
     end
   end
-
 end
--- a/app/services/manual_runner.rb
+++ b/app/services/manual_runner.rb
@@ -7,7 +7,7 @@ class ManualRunner
 
   def run_commits_since(date)
     fetched_commits = fetch_commits_since(date)
-    formated_commits = format_commits(fetched_commits)
+    formatted_commits = format_commits(fetched_commits)
 
     CommitsRunner.run(formated_commits)
   end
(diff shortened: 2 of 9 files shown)


tgxworld:
Excellent work :) This is what I've been wanting to do but have not gotten around to do so. This can be a seperate PR but can we add an admin interface as well?

app/services/commits_runner.rb
```diff
@@ -0,0 +1,25 @@
+module CommitsRunner
```
tgxworld: Let's just make this a class :) I usually prefer to use modules only if you're including them into other classes.

app/services/commits_runner.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+    end
+  end
+
+  def valid?(commit)
+    !Commit.merge_or_skip_ci?(commit[:message]) && Commit.valid_author?(commit[:author_name])
+  end
+
+  def create(commit)
+    Commit.find_or_create_by(sha1: commit[:sha]) do |c|
```
tgxworld: I think we might need to check if the creation is successfully. Otherwise, it might silently fail.

app/services/manual_runner.rb
```diff
@@ -0,0 +1,39 @@
...
+    fetched_commits = fetch_commits_since(date)
+    formated_commits = format_commits(fetched_commits)
+
+    CommitsRunner.run(formated_commits)
+  end
+
+  private
+
+  def fetch_commits_since(date)
+    Octokit.auto_paginate = true
```
tgxworld: Does this paginate the response? Otherwise we might be fetching 2000 commits into memory.

test/services/commits_runner_test.rb
```diff
@@ -0,0 +1,43 @@
+require 'test_helper'
+
+class CommitsRunnerTest < ActiveSupport::TestCase
+  setup do
+    @repo = create(:repo)
+    BenchmarkPool.stubs(:enqueue)
```
tgxworld: Instead of stubbing, perhaps we can look into the using the helps provided by sidekiq? 

https://github.com/mperham/sidekiq/wiki/Testing

bmarkons:
@tgxworld Yes, I was thinking about some admin UI, but wasn't sure. We can add it in later PR :clap: 

I would rather keep this PR separate and use ManualRunner from console to run sequel last 2000 commits.

test/services/commits_runner_test.rb
bmarkons: sure :+1:

app/services/manual_runner.rb
bmarkons: Yeah, this will cause 2000 commits to be fetched into memory.

Maybe I could disable `auto_paginate` and enqueue commits after each fetched page instead of fetching all and enqueuing them all after?

app/services/commits_runner.rb
bmarkons: :+1: will switch to `find_or_create_by!`

bmarkons:
Please feel free to drop suggestions on PR if you have any. Once we have this merged, I will execute benchmarks for last 2000 commits for sequel repo.

BTW, we already have some sequel runs present https://rubybench.org/jeremyevans/sequel/commits

noahgibbs:
Looks reasonable to me.

app/services/manual_runner.rb
```diff
@@ -0,0 +1,46 @@
...
+        repo: {
+          id: @repo.id,
+          name: @repo.name
+        },
+        url: commit['html_url'],
+        created_at: commit['commit']['author']['date'],
+        author: {
+          name: commit['commit']['author']['name']
+        }
+      }
```
robin850: Given the amount of data that this method will have to handle, it may be worth trying to optimize it as much as you can. This is certainly going to be faster using something like:

~~~ruby
def format_commits(commits)
  commits.each do |commit|
    commit[:repo] = @repo
    commit.merge(commit['commit'])
    commit.symbolize_keys!
  end

  commits
end
~~~

You don't really care here if the objects are modified here. Also with the above example you don't have exactly the same API but not sure it's worth allocating a new hash to be able to do something like `commit[:url]` instead of `commit[:html_url]`.

I may be wrong though (both on the fact that you can modify objects without copying them and on the fact that may be faster) ; maybe you can write a tiny benchmark with `benchmark/ips` to see how fast it is and also check the number of allocated objects. What do you think ? 🙂

app/services/commits_runner.rb
```diff
@@ -0,0 +1,25 @@
...
+        create(commit)
+        BenchmarkPool.enqueue(commit[:repo][:name], commit[:sha])
+      end
+    end
+  end
+
+  private
+
+  def self.valid?(commit)
+    !Commit.merge_or_skip_ci?(commit[:message]) && Commit.valid_author?(commit[:author_name])
```
robin850: Unless I'm missing something, this will always return `nil`, it should rather be `commit[:author][:name]`.

test/services/commits_runner_test.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    expect_created(commits)
+    assert_enqueued_jobs(commits.count)
+  end
+
+  def expect_created(commits_hashes)
+    commits_hashes.each do |hash|
+      commit = Commit.find_by!(sha1: hash[:sha])
+
+      assert_equal commit.sha1, hash[:sha]
```
robin850: I guess you should put arguments the other way around to have a non-misleading error message in case things are not equal.

~~~ruby
assert_equal hash[:sha], commit.sha1
~~~

app/services/manual_runner.rb
noahgibbs: If this is running the actual tests for each commit (and it is, right?) then the amount of work in this method will absolutely pale in comparison to the time to run the tests :-(

noahgibbs: 2000 commits (so 2000 times creating a new hash table) just won't be a big deal on that scale.

robin850: Since `:repo` and `:author` references a hash as well, it would rather be 6000 times creating a Hash (while they are already allocated somewhere by Octokit). Either way, your call Marko, I don't mind ; there's certainly no big deal. :-)

app/services/commits_runner.rb
bmarkons: Oh yeah, good catch. I wonder how this passed test then :confused: will fix it.

app/services/manual_runner.rb
bmarkons: Thank you guys for reviews ! :bowing_man: 

> not sure it's worth allocating a new hash


Excuse me for no explanation why do I format commits before passing them to `CommitsRunner`.

It is simply because commits fetched from api and webhook don't have the same format and `CommitsRunner` is intended to handle both manually and automatically (via webhook) fetched commits.

So far, we've only handled commits from webhook via [`GithubEventHandler`](https://github.com/ruby-bench/ruby-bench-web/blob/master/app/services/github_event_handler.rb). Since now we have introduced `ManualRunner` I want to move creating and running commits in `CommitsRunner`. Therefore `GithubEventHandler` and `ManualRunner` should only receive commits and pass them to `CommitsRunner` in specific format.

> Since :repo and :author references a hash as well, it would rather be 6000 times creating a Hash

I wasn't thinking about performance too much here simply because I don't find it very crucial in this scenario as @noahgibbs noticed. But creating 2000 hashes instead of 6000 can be fixed with this:
```
{
  sha: commit['sha'],
  message: commit['commit']['message'],
  repo_id: @repo.id,
  repo_name: @repo.name,
  url: commit['html_url'],
  created_at: commit['commit']['author']['date'],
  author_name: commit['commit']['author']['name']
}
```
I'll change that :+1:

bmarkons: I plan to move creating and running commits from GithubEventHandler in later PR. :)

robin850: Okay, never mind then, my comment isn't really relevant ! :-)

bmarkons:
Thanks @robin850, I've made suggested changes :relieved:

bmarkons:
@tgxworld can we merge this one? :)

PR #214, opened 2017-06-19
Fix commits chart data

bmarkons:
Commit messages containing ruby like string interpolation syntax were causing problem with chart categories and tooltips.

On ruby side messages are being received escaped `\#{ ... }` which caused error in javascript on `JSON.parse`. So before drawing chart I removed those escapements.

I struggled a bit with this but got help https://stackoverflow.com/questions/44630300/why-jquery-cant-parse-inside-array-string/44630468#44630468


tgxworld:
Thank you for fixing :) Feel free to deploy your changes

PR #216, opened 2017-06-21
Show both scripts below comparing commits graphs

bmarkons:
We need to see both benchmark scripts when comparing :chart_with_downwards_trend:

### DIFF
--- a/test/acceptance/compare_benchmarks_test.rb
+++ b/test/acceptance/compare_benchmarks_test.rb
@@ -72,4 +72,15 @@ class CompareBenchmarks < AcceptanceTest
       end
     end
   end
+
+  test 'User should be able to see both scripts' do
+    visit commits_path(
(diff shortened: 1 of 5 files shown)


test/acceptance/compare_benchmarks_test.rb
```diff
@@ -72,4 +72,15 @@ class CompareBenchmarks < AcceptanceTest
       end
     end
   end
+
+  test 'User should be able to see both scripts' do
+    visit commits_path(
```
tgxworld: 👍  always nice having a test

tgxworld:
Looks good to me :)

test/acceptance/compare_benchmarks_test.rb
bmarkons: Yeah, no doubt :)

PR #225, opened 2017-06-27
Use minitest-rails-capybara for acceptance tests

bmarkons:
Acceptance tests were flaky for some time due to possible race conditions. 

I found out that Capybara shouldn't be used with integration test (`ActionDispatch::IntegrationTest`), therefore I replaced it with `Capybara::Rails::TestCase`.
http://blowmage.com/minitest-rails-capybara/

It looks like this recovered test suite - I've run it 10 times in for loop and no flaky error was raised.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,6 @@ group :test do
   gem 'mocha'
   gem 'vcr'
   gem 'webmock'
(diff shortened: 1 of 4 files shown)


tgxworld:
@bmarkons Nice find :+1:

PR #226, opened 2017-06-30
Rename models [WIP]

bmarkons:
This PR is part of my contributions on stuff I found hard to follow while getting familiar with the code.

I was confused a bit with model naming therefore I suggest these changes:
- BenchmarkType -> Benchmark
- BenchmarkResultType -> ResultType

@tgxworld WDYT?


bmarkons:
It looks like model can't be named Benchmark because of name conflict with ruby [module](https://ruby-doc.org/stdlib-1.9.3/libdoc/benchmark/rdoc/Benchmark.html), which is required by default. :man_facepalming:

https://stackoverflow.com/questions/44903703/how-to-avoid-module-conflicts-in-rails

tgxworld:
Hmm I think `BenchmarkResultType` sounds OK though.

bmarkons:
@tgxworld at first I was mainly confused with having two kind of types BenchmarkType and BenchmarkResultType.

Though I wasn't quite sure in naming I've proposed here, so I'll discard it.

Thanks for feedback :bowing_man:

PR #227, opened 2017-07-05
Add admin UI

bmarkons:
I used this bootstrap theme :
https://startbootstrap.com/template-overviews/sb-admin-2

I made new admin layout which is used by AdminController and reorganized assets to not have them mixed.

This PR goal is to make manual run possible through admin UI.


app/controllers/admin_controller.rb
```diff
@@ -2,10 +2,33 @@ class AdminController < ApplicationController
...
+  end
+
+  def run
+  end
+
+  private
+
+  def set_repos
+    @repos = Repo.all
+  end
```
tgxworld: Missing blank line after this line :)

app/controllers/admin_controller.rb
```diff
@@ -2,10 +2,33 @@ class AdminController < ApplicationController
...
+  def set_repo
+    @repo = Repo.find_by_name(params[:repo_name])
```
tgxworld: Probably best to use `find_by(name: )` here since `find_by_*` is deprecated in rails 5

app/controllers/admin_controller.rb
bmarkons: :ok_hand:

bmarkons:
Manual run is now possible through Admin UI :rocket:

.travis.yml
```diff
@@ -13,6 +13,7 @@ bundler_args: --without development:production --deployment --retry=3 --jobs=3
 before_script:
   - bundle exec rake db:create
   - bundle exec rake db:migrate
+  - bundle exec rake assets:precompile &> /dev/null
```
tgxworld: Hmm this is alittle odd. Why do we need to precompile assets in our test env?

app/views/admin/repo.html.haml
```diff
@@ -0,0 +1,22 @@
+- content_for(:page_header) do
+  = "#{@repo.name.capitalize} repo"
+
+
```
tgxworld: Style: Alot of blanks lines here  :)

app/views/admin/home.html.haml
```diff
@@ -0,0 +1,2 @@
+- content_for(:page_header) do
+  = "Dashboard"
```
tgxworld: One good practice is to always use `i18n` when displaying text. You never know when we might need to support other languages in the future :)

test/acceptance/admin_test.rb
```diff
@@ -0,0 +1,10 @@
+require 'acceptance/test_helper'
+
+class AdminTest < AcceptanceTest
+  test 'Admin should be able to manually run suite for existing repo' do
+    repo = create(:repo)
+
+    visit admin_repo_path(repo.name)
+    page.must_have_content 'Run last'
```
tgxworld: Using `i18n` for text that are displayed will also allow you to avoid duplications in the test cases.

.travis.yml
bmarkons: > According to [teampoltergeist/poltergeist#677 (comment)](https://github.com/teampoltergeist/poltergeist/issues/677#issuecomment-222919584), a possible cause of the "Request failed to reach server" problem is that assets are being compiled on-demand, causing enough of a slowdown for PhantomJS to think the connection has failed. This turns off lazy asset compilation in the test environment and precompiles assets before testing on Travis.

I encountered this [problem on Travis](https://travis-ci.org/ruby-bench/ruby-bench-web/builds/250922521#L622) and solved it with assets precompilation.

Error occurred most probably because of introducing admin layout with its own asset group.

app/views/admin/home.html.haml
bmarkons: oh, sure.. forgot about `i18n`:confused: I'll change this :+1:

app/views/admin/repo.html.haml
bmarkons: :man_facepalming:

tgxworld:
OK :+1: LGTM merge away and we can test it out in production :) Good work :+1:

Issue #230, opened 2017-07-14
Improve comparison chart


bmarkons:
Chart isn't displayed nicely when there is not enough results to show.
https://community.rubybench.org/t/bug-in-benchmark-comparison/147

I am working toward this type of chart:
https://www.highcharts.com/stock/demo/compare

PR #231, opened 2017-07-17
Improve comparison chart

bmarkons:
Work regarding https://github.com/ruby-bench/ruby-bench-web/issues/230 :hammer_and_wrench:


bmarkons:
image

app/controllers/repos_controller.rb
```diff
@@ -98,53 +96,69 @@ def benchmark_runs_for(benchmark_type)
...
+          datasets: chart.columns,
+          versions: chart.categories
+        }.to_msgpack
+      )
+    end
+  end
+
+  def cache_comparison(chart, benchmark_type)
     $redis.set(
       @cache_keys[benchmark_type],
```
tgxworld: Hmm don't we have to change the key here? Otherwise it'll override the cache for displaying charts.

bmarkons: It will override cache only for that particular comparison chart. I am setting up cache keys in before_filter.

app/services/comparison_chart_builder.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    [xmin, xmax]
+  end
+
+  def build_data_for(benchmark_type)
+    commit_urls = []
+    series_data = []
+
+    BenchmarkRun.fetch_commit_benchmark_runs(benchmark_type.category, @benchmark_result_type, nil)
+    .sort_by { |run| run.initiator.created_at }
```
tgxworld: Style: Please indent this so it is clear that the method is called on the line above

bmarkons:
May I merge this one?

This PR only affects comparison chart.

tgxworld:
Thank you @bmarkons, looking good to me.

Issue #232, opened 2017-07-20
Make comparison groups for benchmarks


bmarkons:
We need to restrict making comparison graphs only for benchmarks for which it makes sense to be compared.

The most convenient way to achieve this is through admin UI by marking benchmarks with groups. So we would be able to select candidates for comparison from same groups only.

PR #234, opened 2017-07-20
Grouping benchmarks

bmarkons:
Work regarding https://github.com/ruby-bench/ruby-bench-web/issues/232


app/views/admin/groups/_form.html.haml
```diff
@@ -0,0 +1,31 @@
+= form_for(group, url: admin_groups_path(group)) do |f|
+  - if group.errors.any?
+    .alert.alert-danger
+      %h2
+        = "#{pluralize(group.errors.count, "error")} prohibited this admin_group from being saved:"
```
robin850: s/admin_group/group ?

app/helpers/admin/groups_helper.rb
```diff
@@ -0,0 +1,54 @@
+module Admin::GroupsHelper
+  def group_panel_html(group)
+    "
+    <div class='col-lg-6'>
```
bmarkons: :+1: definitely

app/helpers/admin/groups_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+    group.benchmark_types.each do |benchmark_type|
+      benchmark_list << "<li>#{benchmark_type.category}</li>\n"
+    end
+
+    benchmark_list
+  end
+
+  def edit_link(group)
+    link_to(edit_admin_group_path(group), class: 'btn btn-default btn-circle') do
+      '<i class="fa fa-pencil"></i>'.html_safe
```
bmarkons: I was looking for this but without success. Thanks @robin850!

bmarkons:
According to me, this one is ready to :ship: 

Please feel free to drop suggestions :)
@robin850 @noahgibbs

noahgibbs:
Looks good to me.

bmarkons:
@tgxworld May I merge this one?

app/controllers/admin/groups_controller.rb
```diff
@@ -0,0 +1,63 @@
+class Admin::GroupsController < ApplicationController
```
robin850: Sorry, a bit late but why don't you rename the current `admin` controller to `admin/index` and create an `AdminController` class that would encapsulate the authentication logic, `set_admin`, etc. ? 

Maybe in another PR, I don't mind, just asking what you think about this. :-)

bmarkons: You are totally right. All admin controllers should inherit AdminController. This way I have duplicated stuff in AdminController and Admin::GroupsController.

> Maybe in another PR, I don't mind, just asking what you think about this. :-)

I wanted to do it in next PR, do you think it would be more proper doing it in this one?

robin850: It doesn't really matter, it will be in a separate commit anyway so it's up to you ! :-)

PR #237, opened 2017-07-27
Deploy grouping benchmarks

bmarkons:


app/views/admin/groups/_form.html.haml
```diff
@@ -0,0 +1,29 @@
...
+
+  .form-group
+    = f.label :name
+    = f.text_field :name, class: 'form-control'
+
+  .form-group
+    = f.label :description
+    = f.text_area :description, class: 'form-control'
+
+  - Repo.all.each do |repo|
```
noahgibbs: Should repos be passed from the controller? I always feel a bit funny about database queries in the view...

bmarkons: I missed that. Thanks :)

PR #241, opened 2017-08-14
Manually run only latest commit

bmarkons:
There was a bug when running less than 100 latest commits. It was because of paginating 100 commits per page.

Now it's possible to run only latest or last 20 or 50 commits.

### DIFF
--- a/app/services/commits_runner.rb
+++ b/app/services/commits_runner.rb
@@ -1,16 +1,8 @@
 class CommitsRunner
   def self.run(commits, pattern = '')
-    commits.each do |commit|
-      if valid?(commit)
-        create(commit)
-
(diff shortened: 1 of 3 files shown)


noahgibbs:
I'm looking at the manual runner test... Do you ever run a test with fewer commits, less than a full page?

bmarkons:
Yes, here is the case : https://github.com/ruby-bench/ruby-bench-web/pull/241/files#diff-9e0630410bdf7c8a8030021b3a66c251R38

Sometimes I simply want to run only latest to check if setup works properly :)

noahgibbs:
Ah, okay. I misread the pagination size on that. Awesome, thanks!

PR #252, opened 2017-08-24
Update setup section in README

bmarkons:


README.md
```diff
@@ -41,18 +41,18 @@ Motherboard | Intel Coporation S1200RP
...
+- PhantomJS
 
 ## Setup
 ```bash
 redis-server
-bundle exec rake db:setup
 
-# copy database.yml.example to database.yml and make sure you have correct configuration
-cp config/database.yml.example config/database.yml
+# Make sure user with username you use exists in postgres
```
robin850: What do you think about "Make sure that the current user has a Postgres account" ?

bmarkons: Totally :+1:

noahgibbs:
Useful! LGTM.

PR #256, opened 2017-09-07
Fix redis cache deletion

bmarkons:
When new run is being saved for certain benchmark, we want to invalidate cache for both comparison charts.

For example -- when new run for `pg/scope_all` is received, we want to invalidate cache for both charts:

https://rubybench.org/ged/ruby-pg/commits?result_type=pg/scope_all&display_count=2000&compare_with=activerecord/postgres_scope_all

and

https://rubybench.org/rails/rails/commits?result_type=activerecord/postgres_scope_all&display_count=2000&compare_with=pg/scope_all


noahgibbs:
LGTM

PR #259, opened 2018-06-26
Fix chart builder logic for JIT

k0kubun:
## Problem
When using "Show the last 2000 Results", the JIT graph appears only in the commits which didn't have JIT.

screenshot from 2018-06-27 00-32-42
https://rubybench.org/ruby/ruby/commits?result_type=app_erb&display_count=2000

## Solution
Fill `nil` for the commits which don't have `*_jit` result in the chart.

### DIFF
--- a/app/services/chart_builder.rb
+++ b/app/services/chart_builder.rb
@@ -37,6 +37,7 @@ class ChartBuilder
   end
 
   def build_columns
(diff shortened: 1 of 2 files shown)


tgxworld:
Thank you @k0kubun :+1:

PR #280, opened 2020-06-08
Make overview page display one data point per week

OsamaSayegh:
Currently the overview page (e.g. `/ruby/ruby/commits/overview`) tries to display a graph with 2000 data points (`BenchmarkRun`) for every `BenchmarkType`. This is very expensive and takes the page about 5 _minutes_ to load on my local instance.

This PR changes the overview page so it picks a single commit per week and displays the data point associated with that commit. The page now takes 1.5 seconds to load locally without cache. We already have a 30 minutes cache in place for the page, so we don't have to do the work repeatedly.

Screenshot:

image

I also added a link to the overview page on the `/benchmarks` page:

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ vendor/bundle
 
 *.swp
 *.swo
+.ruby-version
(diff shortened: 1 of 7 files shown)


OsamaSayegh:
@tgxworld can you review this whenever you get a chance? thanks! ^_^

